PR #172, opened 2014-09-08
Fix Ctrl-C on some Windows versions, attempt to delete profile multiple times

### LINKED ISSUE #120: Ctrl-C doesn't terminate the Firefox process spawned by jpm run
When using cfx run, using Ctrl-C in the terminal would end both cfx and the Firefox process it spawned.
Using Ctrl-C when jpm is in progress only ends jpm, not it's associated Firefox process.

XrXr:
Closes #120 

Interestingly, #120 doesn't occur on my Windows 8.1 system, but does on Windows 7

### DIFF
--- a/lib/firefox.js
+++ b/lib/firefox.js
@@ -5,6 +5,7 @@ var createProfile = require("./profile");
 var console = require("./utils").console;
 var normalizeBinary = require("./utils").normalizeBinary;
 var getID = require("jetpack-id");
+var fs = require("fs-extra");
 var TEST_RESULTS_REGEX = /\d+ of \d+ tests passed/i
 var isTest = process.env.NODE_ENV === "test";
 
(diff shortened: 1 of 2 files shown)


lib/firefox.js
```diff
@@ -113,6 +114,17 @@ function runFirefox (manifest, options) {
 
     function killFirefox () {
       firefox.kill();
+      var tries = 0;
+      // Windows release lock on files moments after the process ends
+      while (fs.existsSync(profile) && tries < 1000) {
+        try {
+          fs.removeSync(profile);
```
erikvold: if this were successful then we shouldn't try again.

lib/firefox.js
```diff
@@ -113,6 +114,17 @@ function runFirefox (manifest, options) {
 
     function killFirefox () {
       firefox.kill();
+      var tries = 0;
+      // Windows release lock on files moments after the process ends
+      while (fs.existsSync(profile) && tries < 1000) {
```
erikvold: it looks like this is not guaranteed to work any better than things do now.  We should see if there is some alternative, like a package on npm which will handle removing files across platforms well.

erikvold:
I think we should try to use an external module for removing this folder if it requires extra attention, though [it appears the `firefox-profile-js` should be handling this for us](https://github.com/saadtazi/firefox-profile-js/blob/1dc79742c0b61ad2cb3ec7a4b3552672ea9e2073/lib/firefox_profile.js#L134-L137), or at least it should, so we should open these bugs there should we not?

PR #212, opened 2014-10-29
Include icons in xpi base path (fixes #197)

### LINKED ISSUE #197: Relative icon paths do not work
cfx includes icon.png and icon64.png in the base directory of the Add-on, as described [here](https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#iconURL).

jpm does not include icon.png and icon64.png, but instead adds them to the install.rdf, which is unnecessary.

johannhof:
This fixes icons not showing when they were not placed as icon.png in the base directory (fixing #197). I tried to do the same thing cfx does, automatically including the icon file in the xpi root.

Also some tests! :tada: 

Even though I do it here, I'm of the general opinion that we should be careful with creating more temporary files and cleaning them up after the xpi is created. This can only be done to a certain extent, as it comes with a lot of overhead (checking if the file exists, testing that we do not overwrite/delete the file) that could be avoided by using a tmp location.

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -42,6 +42,8 @@ function createRDF (manifest) {
     "em:version": manifest.version || "0.0.0",
     "em:name": manifest.title || manifest.name || "Untitled",
     "em:description": manifest.description || "",
+    "em:iconURL": manifest.icon,
+    "em:icon64URL": manifest.icon64,
     "em:creator": formatAuthor(manifest.author)
   };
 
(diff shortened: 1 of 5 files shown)


johannhof:
Ok, good :bowtie: 
That's kind of what I was trying to say in my initial comment. I dislike the file moving/temporary files but I was trying to get as close to cfx as possible. It's good to have a decision on that. I'll ask more closely next time.

I'll extract the non-moving parts and make a new PR. :)

Issue #178, opened 2014-09-23
Default Firefox path for jpm doesn't work on Ubuntu


ericschultz:
I'm running Ubuntu 14.04 with Firefox 34.0a2 (2014-09-22)
----
While trying to run 'jpm test' on a new add-in created with 'jpm init,' I receive the following output (minus formatting and JPM header at the beginning):

```bash
Running test on Desktop
The `version` entry must be specified, and it should follow semantic
versioning rules.

http://semver.org

XPI created at /tmp/jid1-5RUIvXqVSOhnMw@jetpack.xpi (13605ms)
Error: spawn ENOENT
[object Object]
```
I couldn't figure out what was going on at first. I eventually ran 'jpm test -v' and noticed something: the path being used by jpm test was /usr/lib64/firefox. /usr/lib64/firefox is not where the Firefox binary is on Ubuntu; instead it's in /usr/bin/firefox (might be symlinked elsewhere too).

Ideally jpm should figure out where the firefox installation is on the system and use that for tests. How we could handle cases where there are multiple versions installed is not entirely clear to me.

As a mitigation, it would be nice if jpm checked for the firefox installation and provided a useful error message so people would know what the problem actually is. I just stumbled on this by running jpm test -v but that shouldn't be necessary.

chocolateboy:
`/usr/bin/firefox` isn't a binary:

```sh
file /usr/bin/firefox
/usr/bin/firefox: symbolic link to `../lib/firefox/firefox.sh'
```

The path resolution code in `normalizeBinary` (`lib/utils.js`) looks like it's meant to probe (e.g.):

* "firefox on linux": "/usr/lib/firefox"
* "firefox on linux(64)": "/usr/lib64/firefox"

but instead probes:

* "firefox on linux(64)": "/usr/lib64/firefox"
* "firefox on linux(64)(64)": undefined

But even the "correct" paths don't work: they're directories. On my system (Ubuntu 14.04), the binary is in: `/usr/lib/firefox/firefox`. The only reference to any of the other paths I can find (@shosca) is: [`/usr/lib/firefox-nightly/firefox-nightly`](https://github.com/shosca/-pkgbuild-mozilla-git/blob/master/firefox-nightly/firefox-nightly.desktop).

Note: `arch` is still being appended to `platform` on Windows even though it isn't used (the lookup table is no longer being used for Windows paths). Someone else can clean that up :-)

chocolateboy:
[...]

PR #240, opened 2015-01-09
binary path resolution: cleanup + Linux fix (use `which`)

chocolateboy:
Another attempt at a fix (+ cleanup) for #178. Uses [`which`](https://github.com/isaacs/node-which) to resolve the path on Linux &c.

The tests pass for all platforms, but it has only been tested manually on Linux (Ubuntu 14.04).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -10,7 +10,11 @@ var Winreg = require('winreg');
 var when = require("when");
 var parse = require("mozilla-toolkit-versioning").parse;
 var compare = require("mozilla-version-comparator");
+var nodefn = require("when/node");
+var which = nodefn.lift(require("which"));
+
 var AOM_SUPPORT_VERSION = require("./settings").AOM_SUPPORT_VERSION;
+var APP_NAMES = ["firefox", "beta", "nightly", "aurora"];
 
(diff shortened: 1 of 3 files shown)


chocolateboy:
Note: the [test failure](https://travis-ci.org/mozilla/jpm/jobs/46398287) appears to be unrelated to this patch. The [second run](https://travis-ci.org/mozilla/jpm/jobs/46398288) passes without errors.

erikvold:
hey @chocolateboy I'm sorry but I've moved all of this code to a separate module [`erikvold/fx-runner`](https://github.com/erikvold/fx-runner) could you make a pull request there?  also it appears Firefox release and Firefox beta will require signing soon, so we can pretty much remove all of the code using those two channels now.

chocolateboy:
Having spent several days writing, rewriting, testing and tweaking this on the understanding that something labeled as a "good first bug" would at least merit prompt feedback, I'm not particuarly motivated to re-invest all that time re-implementing this code in a PR for a completely different repo, where -- presumably -- it will languish again indefinitely. Wouldn't it have made more sense to have applied this patch *before* forking the code?

PR #242, opened 2015-01-12
add name option to jpm init

johannhof:
This enables the jpm user to call `jpm init my-addon-name` and have it initialize in a new directory called "my-addon-name", also the name in the prompt wizard default is set to that name.

cfx already does this, for instance.

Calls to `jpm init` without a name will still initialize in the same directory the user is currently in.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -57,10 +57,10 @@ program
   }));
 
 program
-  .command("init")
+  .command("init [name]")
   .description("Series of prompts to create a package.json for a new addon")
-  .action(function () {
-    init().then(process.exit);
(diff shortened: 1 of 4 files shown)


erikvold:
This differs from `npm init` and I'm not sure it improves the usage.

This makes it possible to do:

press `jpm init foo` enter -> press enter -> ..

Instead of

press `jpm init` enter -> press `foo` enter -> ..

which is no fewer keys to press, so I don't see the benefit.

r- for this reason, and the fact that jpm can be wrapped by another node based cli which can do the requested feature.

PR #267, opened 2015-02-22
Convert relative icon paths to resource urls

johannhof:
This is a temporary hack to workaround #197 until the issue is resolved and landed in Firefox. 

Converting icon paths to resource urls has several drawbacks, such as not displaying when the add-on is turned off and the possibility of resource:// being changed at some point in the future. Hence the warning.

The code for `normalizeID` is mostly taken from https://github.com/mozilla/addon-sdk/blob/12391f89c0fc82fbd4683c87df2aee8dfc58772d/lib/sdk/addon/bootstrap.js#L25

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -7,6 +7,8 @@ var jsontoxml = require("jsontoxml");
 var getID = require("jetpack-id");
 var parse = require("mozilla-toolkit-versioning").parse;
 var GUIDS = require("./settings").ids;
+var utils = require("./utils");
+var console = require("./utils").console;
 var MIN_VERSION = require("./settings").MIN_VERSION;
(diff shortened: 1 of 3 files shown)


erikvold:
So there is a plan to change from `resource:` uris to something like `addon:` uris which would mean any add-ons made with jpm and using relative path mapping like this could break depending on when we implement this, this is [bug 779197](https://bugzilla.mozilla.org/show_bug.cgi?id=779197).  Also using the default paths `$root/icon.png` and `$root/icon64.png` is the best option because those icons will be displayed when the add-on is disabled too, so I think we should recommend/promote this implementation over any other.

I'm starting to think that we should just wait and support this on the m-c side in XPIProvider.jsm.

PR #268, opened 2015-02-22
Dedupe "xpi created" messages

johannhof:
Before:
screenshot 2015-02-22 10 50 09

After:
screenshot 2015-02-22 10 47 27

If you want to show the xpi location in `jpm test` or `jpm run`, you can use the -v flag.

### DIFF
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -8,6 +8,7 @@ var getID = require("jetpack-id");
 var validate = require("./validate");
 var hasAOMSupport = require("./utils").hasAOMSupport;
 var utils = require("./xpi/utils");
+var console = require("./utils").console;
 var _ = require("lodash");
 
 /**
@@ -59,14 +60,18 @@ function xpi(manifest, options) {
     })
     .then(function(fallbacks) {
       options.fallbacks = fallbacks;
-      console.log("Creating XPI");
+      if (options.verbose){
+        console.log("Creating XPI");
+      }
       return utils.createZip(options);
     })
     .then(function() {
       return utils.createUpdateRDF(options, updateRdfPath);
     })
     .then(function() {
-      console.log("Created XPI at " + xpiPath);
+      if (options.verbose){
+        console.log("Created XPI at " + xpiPath);
+      }
       return xpiPath;
     });
 }
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -60,7 +60,7 @@ function xpi(manifest, options) {
     })
     .then(function(fallbacks) {
       options.fallbacks = fallbacks;
-      if (options.verbose){
+      if (options.verbose) {
         console.log("Creating XPI");
       }
       return utils.createZip(options);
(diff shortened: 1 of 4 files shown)


test/functional/test.run.js
```diff
@@ -124,20 +124,9 @@ describe("jpm run", function () {
...
       var proc = exec("run", options, function (err, stdout, stderr) {
         expect(err).to.not.be.ok;
-        expect(stdout).to.contain("console\.log: param-dump:")
+        expect(stdout).to.contain("console\.log: param-dump:");
         expect(stdout).to.contain("PARAMS DUMP START");
-        done();
-      });
-    });
-
-    it("logs out everything from console with -v", function (done) {
```
erikvold: why is this test removed?

johannhof: It never worked correctly, if I'm not mistaken. We strip out all logs in testing and that `expect(stdout.split("\n").length).to.be.gt(2);` just applies to the two `Creating XPI` messages we forgot to log out with our own util.console. If you take that away it receives exactly the same output as the test above.

johannhof:
@jsantell wanna have a look at this one? :) should be easy

johannhof:
@jsantell ping? :blush:

lib/xpi.js
```diff
@@ -55,14 +56,18 @@ function xpi (manifest, options) {
       return utils.createFallbacks(options);
     })
     .then(function() {
-      console.log("Creating XPI");
+      if(options.verbose){
```
jsantell: nit: space between `if (`

lib/xpi.js
```diff
@@ -55,14 +56,18 @@ function xpi (manifest, options) {
...
+        console.log("Creating XPI");
+      }
       return utils.createZip(options);
     })
     .then(function() {
       return utils.removeFallbacks(options);
     })
     .then(function () {
-      console.log("Created XPI at " + xpiPath);
+      if(options.verbose){
```
jsantell: nit: space between `if (`

test/functional/test.run.js
jsantell: Is there another way we can test `-v` in this case?

jsantell:
Sorry for the delay!

test/functional/test.run.js
johannhof: Well the problem is that we strip logs in testing, as I mentioned. So we could remove that but then lots of other test will probably fail. :confused:

This test working is really just an accident because we forgot to log `Creating XPI` correctly. It didn't even check what it was supposed to check.

In the end it's your call, but I would suggest moving the "not stripping logs in tests anymore" into a separate pull request, which would mean no tests for this.

jsantell: I'm fine with a follow up issue if we remove this!

andymckay:
Would it be possible to get the tests working as per #374 before we look at merging this?

johannhof:
@andymckay @kumar303 I recently used jpm again and noticed that the output still looks rather comical. I remembered and updated this PR. Do you think we can get it merged this time? :)

kumar303:
Sorry, I missed this. It looks good and thanks for resurrecting / fixing the PR.

johannhof:
:+1:

PR #272, opened 2015-02-24
Tweaks jpm to turn it into a nodejs library

rpl:
This pull request explores the changes needed to turn jpm into a nodejs library,
which will help its integration into gulp/grunt plugins.

- [x] add a new ```--addon-dir``` jpm option (which defaults to the current work dir), and remove
  ```process.chdir/process.cwd``` usage
- [x] inject fallback ```install.rdf``` and ```bootstrap.js``` files into the generated xpi archive,
  instead of creating temporary files into the addon directory
- [x] changes in the test suite
- [ ] fix fails on iojs (https://travis-ci.org/mozilla/jpm/jobs/51972298)
- [ ] add test case which tests explicitly that ```--addon-dir``` defaults to the ```process.cwd()```
- [ ] add test cases about *jpm as library* usage

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -25,6 +25,7 @@ program
   .option("-o, --overload [path]", "Overloads the built-in Firefox SDK modules with a local copy located at environment variable `JETPACK_ROOT` or `path` if supplied. Used for development on the SDK itself.")
   .option("-f, --filter <pattern>", "--filter=FILENAME[:TESTNAME] only run tests whose filenames match FILENAME and " +
                       "optionally match TESTNAME, both regexps")
+  .option("--addon-dir <path>", "Path of Firefox addon to build.", process.cwd())
   .option("--binary-args <CMDARGS>", "Pass additional arguments into Firefox.")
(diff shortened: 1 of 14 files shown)


neico:
The new post commands aren't covered by the first task where at least postwatch uses process.cwd()

kumar303:
hi @rpl, do you think we should still do this? Or close it?

bwinton:
I'ld still kinda like it, but I don't think we should put a ton of effort into it.  (I'm trying to switch to `web-ext`, which will hopefully have a node API, and I've mostly switched to just running stuff as npm scripts, which means the command-line version is kinda fine anyways… :wink:)

rpl:
@kumar303 @bwinton 

What we needed from this PR has been already splitted in two PR and merged:
- #399 by @jsnajdr (Thanks!!!) 
- and in  #448

I think that jpm is already nicely embeddable when (and if) needed, e.g. here is how it is integrated into a grunt task in my grunt-jpm repo:

- https://github.com/rpl/grunt-jpm/blob/feature/jpm-test-task/tasks/jpm.js

PR #277, opened 2015-02-28
Need en-US locale in MDN URLs

okuryu:
When locale doesn't contains the MDN URL, MDN redirects to the localized page which contains a UA's locale, but there are cases have gone to not found page. It avoids this problem by specify `en-US` locale clearly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,9 +53,9 @@ npm link
 
 ### Documentation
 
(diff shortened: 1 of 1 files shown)


erikvold:
Thanks! I didn't realize the en-US wasn't used when there were no translations.

PR #280, opened 2015-03-05
Update index.js

richgilbank:
Correcting file paths

### DIFF
--- a/data/index.js
+++ b/data/index.js
@@ -1,7 +1,7 @@
 var self = require('sdk/self');
 
-// a dummy function, to show how tests work. 
-// to see how to test this function, look at ../test/test-main.js
+// a dummy function, to show how tests work.
+// to see how to test this function, look at test/test-index.js
 function dummy(text, callback) {
   callback(text);
 }


data/index.js
```diff
@@ -1,7 +1,7 @@
 var self = require('sdk/self');
 
-// a dummy function, to show how tests work. 
-// to see how to test this function, look at ../test/test-main.js
+// a dummy function, to show how tests work.
+// to see how to test this function, look at test/test-index.js
```
erikvold: `test/test-index.js` -> `../test/test-index.js` please

richgilbank: Running `jpm init` seems to create the following structure:

Is this  used outside of init as well?
https://github.com/mozilla/jpm/blob/master/lib/init.js#L56-L57

erikvold: yes the tests are used for `jpm test`, for instance try `jpm init && jpm test`

richgilbank: Right, it's used in the tests, but that doesn't modify the directory structure at any point, so `index.js` is always a sibling of the `test` directory and getting to `test-index.js` would just be `./test/test-index.js`, no? I think I'm missing something here. Happy to change it if you insist, but I don't see any scenario where `test` is up a directory.

erikvold: ah you're right, thanks!

PR #284, opened 2015-03-18
Fix display of log level

clochix:


johannhof:
ping @erikvold on this, it's quite ugly

Test is failing because randomness.

erikvold:
Thanks!

PR #299, opened 2015-05-14
Suggested fix for Bug 1153954

kkapsner:
https://bugzilla.mozilla.org/show_bug.cgi?id=1153954

### DIFF
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -29,7 +29,7 @@ function xpi(manifest, options) {
   options = options || {};
   var addonDir = options.addonDir || process.cwd();
   var xpiVersion = manifest.version ? ("-" + manifest.version) : "";
-  var xpiName = getID(manifest) + xpiVersion + ".xpi";
+  var xpiName = (manifest.name || (getID(manifest) + xpiVersion)) + ".xpi";
   var updateRdfName = getID(manifest) + xpiVersion + ".update.rdf";
   var xpiPath = join(options.xpiPath || addonDir, xpiName);
   var updateRdfPath = join(options.xpiPath || addonDir, updateRdfName);
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -30,15 +30,18 @@ function xpi(manifest, options) {
   var addonDir = options.addonDir || process.cwd();
   var xpiVersion = manifest.version ? ("-" + manifest.version) : "";
   var xpiName;
-  if (manifest.name){
-    if (manifest.name.match(/^[a-z0-9_\-\.]+$/i)){
+  if (manifest.name) {
+    if (manifest.name.match(/^[a-z0-9_\-\.]+$/i)) {
       xpiName = manifest.name;
-    }
-    else {
-      console.warn("The provided addon name \"" + manifest.name + "\" is not a valid filename. Using ID instead.");
+    } else {
+      console.warn(
+        "The provided addon name \"" +
+        manifest.name + 
+        "\" is not a valid filename. Using ID instead."
+      );
     }
   }
-  if (!xpiName){
+  if (!xpiName) {
     xpiName = getID(manifest) + xpiVersion;
   }
   xpiName += ".xpi";
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -31,12 +31,14 @@ function xpi(manifest, options) {
   var xpiVersion = manifest.version ? ("-" + manifest.version) : "";
   var xpiName;
   if (manifest.name) {
-    if (manifest.name.match(/^[a-z0-9_\-\.]+$/i)) {
+    // checking that the name only contains save characters for any file system:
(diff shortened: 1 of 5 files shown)


kumar303:
Hi! You submitted this at a time when the project wasn't maintained. If you'd still like to submit this patch, please rebase on master.

kkapsner:
Rebased and it should work now.

kumar303:
Thanks for following up. I took a closer look and I think it's the right change to make but it would need a few other things:
* the name will need to be validated to not have spaces (and other odd characters) since we can't trust developers to always make it such
* the tests will need updating to reflect the new behavior. The easiest way to start is by running `npm test unit` which finishes faster. The integration tests are cumbersome to deal with.

kkapsner:
I added a check for a proper filename and the unit tests are now running without error. But I have no idea why the integration tests are failing...

test/unit/test.xpi.js
```diff
@@ -33,8 +34,10 @@ describe("lib/xpi", function() {
     var xpiPath;
     return xpi(manifest).then(function(filePath) {
       xpiPath = filePath;
+      // expect(xpiPath).to.be.equal(path.join(simpleAddonPath,
+                                            // "@simple-addon-1.0.0.xpi"));
```
johannhof: please delete this, it usually doesn't make sense to keep commented-out code

lib/xpi.js
```diff
@@ -29,7 +29,22 @@ function xpi(manifest, options) {
...
   var xpiVersion = manifest.version ? ("-" + manifest.version) : "";
-  var xpiName = getID(manifest) + xpiVersion + ".xpi";
+  var xpiName;
+  if (manifest.name) {
+    if (manifest.name.match(/^[a-z0-9_\-\.]+$/i)) {
+      xpiName = manifest.name;
+    } else {
+      console.warn(
+        "The provided addon name \"" +
+        manifest.name +
```
johannhof: JSHint is failing your travis build here, because of a trailing whitespace :)

lib/xpi.js
```diff
@@ -29,7 +29,22 @@ function xpi(manifest, options) {
   options = options || {};
   var addonDir = options.addonDir || process.cwd();
   var xpiVersion = manifest.version ? ("-" + manifest.version) : "";
-  var xpiName = getID(manifest) + xpiVersion + ".xpi";
+  var xpiName;
+  if (manifest.name) {
+    if (manifest.name.match(/^[a-z0-9_\-\.]+$/i)) {
```
johannhof: can you add a short comment what and why this is supposed to match?

johannhof:
Hey, your test is failing at test/functional/test.xpi.js, I think you forgot to update the file name there.

While you're at it, please also take a look at my above comments. Thanks for contributing! :)

kkapsner:
All the unit tests pass without problem. But the integration tests fail...

I adressed all your comments mentioned above.

test/unit/test.xpi.js
```diff
@@ -45,13 +46,29 @@ describe("lib/xpi", function() {
     })
     .catch(done);
   });
+  it("Check file name with id", function(done) {
```
kumar303: You don't need to declare `done` here. Just return the promise (as you are) and the test will complete when the promise completes or it will fail when it's rejected.

test/unit/test.xpi.js
```diff
@@ -45,13 +46,29 @@ describe("lib/xpi", function() {
...
+    process.chdir(simpleAddonWithIdPath);
+    var manifest = require(path.join(simpleAddonWithIdPath, "package.json"));
+    var xpiPath;
+    return xpi(manifest).then(function(filePath) {
+      xpiPath = filePath;
+      expect(xpiPath).to.be.equal(path.join(simpleAddonWithIdPath,
+                                            "@simple-addon-1.0.0.xpi"));
+      return true
```
kumar303: missing semi-colon

kumar303: you could do `return xpiPath` and `.then(function(xpiPath) {...})` on the following line to simplify this.

test/unit/test.xpi.js
kkapsner: I just copied the previous test... but I can removed it.

test/unit/test.xpi.js
kkapsner: As like with your other comment - just copied...

kumar303:
The test log shows:

Potentially unhandled rejection [1] Error: ENOENT, open '/home/travis/build/mozilla-jetpack/jpm/test/fixtures/simple-addon/@simple-addon-1.0.0.xpi'

I think something needs to be updated to look for the right file.

kkapsner:
Finally!

johannhof:
Great, thank you!

kkapsner:
You're welcome.

PR #304, opened 2015-06-06
Fix #295: Add "jetpack" keyword by default

### LINKED ISSUE #295: Add "jetpack" keyword by default
I think it makes sense to try and let all add-on sdk _specific_ extensions be findable under one keyword on npmjs.org. It seems "jetpack" has been used so far by most (I am not sure if it covers all modules, as it is hard to catch everything due to different terms being relevant). I think it would make sense, if jpm would by default add "jetpack" as a keyword to the package.json when using jpm init, so it is by default there. Unless this harms the usage of package.json for extension metadata within firefox, of course.

freaktechnik:


jsantell:
Simple enough, thanks @freaktechnik :smile:

PR #305, opened 2015-06-10
Support background option

mdevils:
I want to introduce `--background` option which runs Firefox in background. Currently firefox always gains focus on every `jpm test` which is sometimes very annoying as I loose focus from the current app when `jpm test` runs on change.

With `--background` option I can keep having firefox on background between launches.


DecipherCode:
+1

kumar303:
Hi! You submitted this at a time when the project wasn't maintained. If you'd still like to submit this patch, please rebase on master.

kumar303:
Since I haven't seen a response, I'm closing the pull request for now. Feel free to open a new one against the latest master. Thanks for taking the time to make a patch.

PR #307, opened 2015-06-23
test on node.js 0.12

nikolas:


jsantell:
This was added in c28311cb -- thank you though!

PR #333, opened 2015-09-07
Fix failures in tests and travis builds

rpl:
This PR fixes a number of issues on running tests locally and during travis builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,26 +3,29 @@ language: node_js
 node_js:
   - "iojs"
   - "0.10"
-  - "0.11"
+  - "0.12"
+env:
+  - FIREFOX_VERSION=nightly
+#  - FIREFOX_VERSION=aurora
+#  - FIREFOX_VERSION=beta
+#  - FIREFOX_VERSION=release
 
 notifications:
   irc: "irc.mozilla.org#jpm"
-
-env:
-  - JPM_BUILD_TYPE=travis
 
 before_install:
   - "export DISPLAY=:99.0"
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
 
 before_script:
-  - npm install fx-download -g
+  - npm install firefox-get -g
   - pwd
   - cd ..
-  - fx-download --branch nightly -c prerelease --host ftp.mozilla.org firefox
+  - wget `firefox-get -v $FIREFOX_VERSION` -O firefox.tar.bz2 && tar xvf firefox.tar.bz2
   - cd $TRAVIS_BUILD_DIR
   - pwd
 
 script:
   - export JPM_FIREFOX_BINARY=$TRAVIS_BUILD_DIR/../firefox/firefox
+  - export JPM_BUILD_TYPE=travis
   - npm test
--- a/package.json
+++ b/package.json
@@ -7,12 +7,13 @@
     "jpm": "./bin/jpm"
   },
   "scripts": {
-    "test": "node ./test/run.js",
     "addons": "node ./test/run.js addons",
     "functional": "node ./test/run.js functional",
     "addon_runners": "node ./test/run.js addon_runners",
     "unit": "node ./test/run.js unit",
-    "documentation": "mocha -t 5000 test/documentation"
+    "documentation": "mocha -t 5000 test/documentation",
+    "test": "node ./test/run.js",
+    "prepush": "npm run documentation"
   },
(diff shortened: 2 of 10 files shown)


rpl:
Travis is not currently enabled on this repo, so I've temporarily enabled it on my fork.

Follows the builds of the current master branch and this PR:
- build on this PR (green): https://travis-ci.org/rpl/jpm/builds/79144340
- build on master (red): https://travis-ci.org/rpl/jpm/builds/79127599

rpl:
@jsantell This PR is almost completed and can be reviewed 
(I'm just taking another look to the "firefox download" part, which is a change restricted to the travis config)

jsantell:
@rpl Everything looks good to me! I mentined this in IRC but I think fx-download fails due to using ftp, and if mozilla-download works, great, if not, check out [firefox-get](https://github.com/jsantell/node-firefox-get), which was deprecated for more robust modules, but I think this is the only one that does not use ftp (which was taken down a few weeks ago, hence failures, I believe)

rpl:
@jsantell **mozilla-download** works without problems (it uses the taskcluster api to retrieve the request build artifacts), but I tried **firefox-get**, it works like a charm and I definitely prefer it because we can download the latest build from all the different channels and run a matrix build on travis (and I agree that it would be nice to **un-deprecate** it) 

some of the builds in the matrix are currently failing:

- https://travis-ci.org/rpl/jpm/builds/79319004

apparently for timeout reasons, I'm not currently sure if it is related to temporary issues on travis or because of some of our test cases, I'm going to try to force a rebuild to check it more deeply.

How about re-enabling the travis-ci builds on this repo too?

jsantell:
I reset up the travis hook -- let's see if it's working now?

Let's remove testing for node 0.11 -- that's an unstable version anyway, and we have 0.12 now.

Are these supposed to be pass on non-nightly? I guess it should unless we use latest SDK built into FX for a test or something..

rpl:
@jsantell I've removed the node 0.11 and commented out all the firefox channels besides the nightly one (running tests on all the channels doesn't add any value right now, but it can be useful if in the future we'll introduce changes that need to be tested for retro-compatibility)

unfortunately some of the tests still fail from time to time 
(e.g. the teacher node module sometimes receives an invalid response from the server which it refers to validate the documentation content, and some of the addon tests randomly fail for timeouts)

do we try to resolve the remaining issues in a follow up of this PR?

jsantell:
If the teacher module can't consistently handle testing the spell checking, then I don't think it's even worth having in the tests. Maybe locally, but yeah defeats the purpose of testing.

Intermittents will happen, but if it's passing (it looks like it is), let's merge it!

rpl:
@jsantell yep, I agree completely:
in the last commit I've added test.docs.js in the blacklist, so it will not be included when we run the complete test suite, and tweaked the npm scripts to be able to run it manually.

How about run the "spell checking" tests only on prepublish? 
(if I'm not wrong the "prepublish npm hook" should be run automatically just before the upload of the new package version to npmjs)

and apparently skipping the spell checks during the travis builds contribute to stabilize the random failures too.

jsantell:
I think the prepublish hook sounds like a great idea -- reduces intermittents and still prevents someone from publishing with typos (which isn't a huge deal anyway)

rpl:
@jsantell unfortunately the "prepublish" script runs on "npm install" too (https://docs.npmjs.com/misc/scripts), and "npm install" runs during automated tests... and "teacher" module confirmed that it can't handle the continuous building:
it managed to fail 3 times in a row (because "npm install" retries 3 times before exit with a failure) and turned the run into a failed build (https://travis-ci.org/mozilla-jetpack/jpm/jobs/79819165)

In the last commit I changed it to run on "preversion", but it will be actually useful only if **release-it** bumps the package.json version using the "npm version" command (which will run the preversion script automatically)

rpl:
@jsantell in the last commit I add [husky](https://github.com/typicode/husky) to the npm dev dependencies to run the documentation spellchecks automatically before a "git push".

I think it's a good compromise because on every "git commit" is too annoying (especially during offline work) and we will be sure that the spellchecks run before than merging a branch back to master (and during a "git push"  we are online for sure)

the prepush hook can be disabled with ```--no-verify```

guyzmuch:
Any idea when this is gonna be merge to the master branch? I am looking forward to have it there.

jsantell:
@rpl is this good for a squash and merge? anything else we need?

rpl:
@jsantell PR ready for merge

I've just rebased it on the updated master and squashed it into 2 commits (the former to fix and blacklist the spell-checks test case, the latter to run the spell-checks test case on git push)

NOTE: ```git push --no-verify``` disable the spell-checks test case

PR #361, opened 2015-10-23
Allow a relative path for xpiPath option

FayneAldan:
Makes things a lot easier. ;)


FayneAldan:
Okay. It seems I'm crazy... :/

Issue #327, opened 2015-08-24
Add jpm sign command


DavidBruant:
To sign the addon from the command line.

kumar303:
I'm going to take a stab at a patch for this. Stay tuned...

kumar303:
The signing API itself is getting tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=1206158

kumar303:
patch in progress: https://github.com/mozilla-jetpack/jpm/pull/358

kumar303:
the sign command has been completed and is available in 1.0.4 of jpm. Documentation: https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/jpm#jpm_sign

PR #362, opened 2015-10-26
Wait for validation and show signed XPI

kumar303:
This supports #327

This does not yet download the signed XPI when validation is finished. That will come in another patch.

### DIFF
--- a/lib/amo-client.js
+++ b/lib/amo-client.js
@@ -3,26 +3,38 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
+var _ = require('lodash');
 var fs = require('fs');
 var jwt = require('jsonwebtoken');
 var request = require('request');
 var when = require("when");
 var nodefn = require('when/node');
+var logger = require("./utils").console;
 
 
 /**
  * Construct a new addons.mozilla.org API client.
  *
  * @param {Object} conf
- *   - `apiKey` API key string from the Developer Hub.
- *   - `apiSecret` API secret string from the Developer Hub.
- *   - `apiUrlPrefix` API URL prefix, including any leading paths.
+ *   - `apiKey`: API key string from the Developer Hub.
+ *   - `apiSecret`: API secret string from the Developer Hub.
+ *   - `apiUrlPrefix`: API URL prefix, including any leading paths.
+ *   - `signedStatusCheckInterval`: A period in millesconds between
+ *     checks when waiting on add-on signing.
  */
 function Client(conf) {
-  conf = conf || {};
+  conf = _.assign({
+    signedStatusCheckInterval: 1000,
+  }, conf);
   this.apiKey = conf.apiKey;
   this.apiSecret = conf.apiSecret;
-  this.apiUrlPrefix = conf.apiUrlPrefix;
+  this.apiUrlPrefix = conf.apiUrlPrefix;  // default set in CLI options.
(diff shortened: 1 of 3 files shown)


lib/amo-client.js
```diff
@@ -37,25 +49,98 @@ function Client(conf) {
...
+
+  return when.promise(function(resolve, reject) {
+    self._validateProgress.animate();
+    var statusCheckTimeout;
+    var nextStatusCheck;
+
+    function checkSignedStatus() {
+      self.get({url: statusUrl}).then(function(result) {
+        var data = result[1];
+        if (data.valid && data.download_url) {
```
mstriemer: This looks like it will keep checking until it passes validation which I think would be forever if it fails validation.

You should be able to check `data.processed` to know when automated validation has been completed. Then `data.valid` should let you know if it passed automated validation. Once it is valid it will get reviewed (which is automatic but still done asynchronously for unlisted) and then `data.reviewed` should become `tru

 If it has passed validation then `data.active` should tell you if the version has been created and signed.

mstriemer: Sent that comment too early. It is now updated.

I think this should be something like:

```js
if (data.processed) {
  if (data.valid) {
    // Looks good. If this is listed we're done (perhaps something in the API should reflect this).
    // If this is unlisted then check that `data.active` is true and show the download URL.
  } else {
    // It failed validation, print out `data.validation`.
  }
} else {
  // Still processing, keep polling.
}

File: lib/amo-client.js [hunk 2]
```diff
@@ -37,25 +49,98 @@ function Client(conf) {
...
+              checkSignedStatus, self.signedStatusCheckInterval);
+        }
+
+        // TODO: show some validation warnings on success.
+        // We should show things like 'missing update URL in install.rdf'
+
+        // TODO: show a URL to validation results when processing
+        // is done and the add-on validation failed.
+        // Something like:
+        // hostFromApiURL + '/developers/upload/' + response.id
```
mstriemer: The URL is in the response now as `validation_url`.

lib/amo-client.js
```diff
@@ -37,25 +49,98 @@ function Client(conf) {
...
+
+    function checkSignedStatus() {
+      self.get({url: statusUrl}).then(function(result) {
+        var data = result[1];
+        if (data.valid && data.download_url) {
+
+          self._validateProgress.finish();
+          logger.log('SUCCESS');
+          // TODO: actually download the signed file from the URL.
+          logger.log('Your add-on has been signed:', data.download_url);
```
mstriemer: `download_url` is actually an array right now. Originally I thought it would be a single URL but there can be multiple files (one for each platform) so they'd each get their own download URL.

An array might not handle multiple files well though since there is some metadata that the user probably wants (what platform it's for at least). So then we could have an array of files or a hash of files keyed off of platform. I'm not sure what is best but `download_url` definitely won't work forever. The API only supports the "ALL" platform right now so maybe I should go back to just `download_url` until we add support.

lib/amo-client.js
kumar303: ah, whoops! I totally forgot to add this, thanks

lib/amo-client.js
```diff
@@ -66,8 +151,9 @@ Client.prototype.sign = function(conf) {
  * @param {Object} conf as accepted by the `request` module.
```
mstriemer: Is there another argument here now that it switched to using `arguments`?

kumar303: yep, fixed

PR #365, opened 2015-10-28
Download signed add-on files

kumar303:
Supports #327

### DIFF
--- a/lib/amo-client.js
+++ b/lib/amo-client.js
@@ -5,6 +5,8 @@
 
 var _ = require('lodash');
 var fs = require('fs');
+var url = require('url');
+var path = require('path');
 var jwt = require('jsonwebtoken');
 var request = require('request');
 var when = require("when");
@@ -118,12 +120,7 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
             // TODO: show some validation warnings if there are any.
             // We should show things like 'missing update URL in install.rdf'
 
-            logger.log('Your add-on has been signed:');
-            // TODO: actually download the signed files.
-            data.files.forEach(function(file) {
-              logger.log('    ' + file.download_url);
-            });
-            resolve({success: true});
+            resolve(self.downloadSignedFiles(data.files));
 
           } else {
             logger.log('Your add-on failed validation and could not be signed');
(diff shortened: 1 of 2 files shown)


lib/amo-client.js
```diff
@@ -151,6 +148,75 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
 
+Client.prototype.downloadSignedFiles = function(signedFiles, options) {
+  options = _.assign({
+    createWriteStream: fs.createWriteStream,
+    request: this._request,
+    stdout: process.stdout,
+  }, options);
+  var self = this;
+  var allDownloads = [];
+  var dataExpected = 0.1;
```
mstriemer: Why do these start with `0.1`? I assumed it would be to avoid dividing by zero when rendering the downloaded percentage but won't this mean it says `Downloading: 100%` and then when it gets the response it changes to `Downloading: 0%`? Maybe I'm missing something but that seems like it is worth handling.

kumar303: yeah, I was trying to avoid divide by zero without an extra if statement. Hmmm.

kumar303: fixed

lib/amo-client.js
```diff
@@ -151,6 +148,76 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
+  options = _.assign({
+    createWriteStream: fs.createWriteStream,
+    request: this._request,
+    stdout: process.stdout,
+  }, options);
+  var self = this;
+  var allDownloads = [];
+  // Set some defaults so it looks like 0% before starting the download.
+  var dataExpected = 100;
+  var dataReceived = 0.1;
```
mstriemer: I think I'd rather see the `if` and `Waiting for download to start...` before it starts. I think you could set `dataExpected = 0.00001` and `dataReceived = 0` if you don't want to do that though.

lib/amo-client.js
```diff
@@ -151,6 +148,76 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
+
+  showProgress();
+  signedFiles.forEach(function(file) {
+    allDownloads.push(download(file.download_url));
+  });
+
+  return when.all(allDownloads).then(function(downloadedFiles) {
+    logger.log('Your signed add-on awaits:');
+    downloadedFiles.forEach(function(fileName) {
+      logger.log('    ' + fileName.replace(process.cwd(), '.'));
```
mstriemer: The replace doesn't look super safe but it's probably fine. Is there some relative path helper you could use instead?

mstriemer:
r+wc

Looks like the build is broken, not sure if that's new or not.

kumar303:
yeah, the tests have always been flaky because of timing issues. I'm hoping to try fixing those soon.

PR #366, opened 2015-10-29
Signing waits for fully validated files (bug fix)

kumar303:
Supports #327

### DIFF
--- a/lib/amo-client.js
+++ b/lib/amo-client.js
@@ -68,18 +68,18 @@ Client.prototype.sign = function(conf) {
     var httpResponse = responseResult[0];
     var response = responseResult[1];
 
-    var acceptableStatuses = [200];
+    var acceptableStatuses = [200, 201, 202];
     if (acceptableStatuses.indexOf(httpResponse.statusCode) === -1) {
       if (typeof response === 'object' && response.error) {
         logger.error('Server response:', response.error,
                      '( status:', httpResponse.statusCode, ')');
         return {success: false};
       }
+
       throw new Error(
         'Received bad response from the server while signing; ' +
         'status: ' + httpResponse.statusCode + '; ' +
-        'response: ' +
-        (typeof response === 'string' ? response.substring(0, 100) + '...' : response));
+        'response: ' + formatResponse(response));
     }
 
     return self.waitForSignedAddon(addonUrl);
@@ -101,6 +101,7 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
     abortAfter: 45000,
   }, opt);
   var self = this;
+  var lastStatusResponse;
 
   return when.promise(function(resolve, reject) {
     self._validateProgress.animate();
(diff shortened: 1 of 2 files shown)


lib/amo-client.js
```diff
@@ -211,13 +219,23 @@ Client.prototype.downloadSignedFiles = function(signedFiles, options) {
     });
   }
 
-  showProgress();
-  signedFiles.forEach(function(file) {
-    allDownloads.push(download(file.download_url));
-  });
+  // TODO: handle 404 downloads
```
kumar303: I'll file a separate issue for this but it probably only impacts local development

lib/amo-client.js
```diff
@@ -68,18 +68,18 @@ Client.prototype.sign = function(conf) {
     var httpResponse = responseResult[0];
     var response = responseResult[1];
 
-    var acceptableStatuses = [200];
+    var acceptableStatuses = [200, 202];
```
mstriemer: I'm currently in the process of adding `201` for when an add-on is created.

lib/amo-client.js
```diff
@@ -110,7 +110,12 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
     function checkSignedStatus() {
       self.get({url: statusUrl}).then(function(result) {
         var data = result[1];
-        if (data.processed) {
+
+        var failedValidation = !data.valid;
+        // The add-on passes validation and all files have been created.
+        var passedValidation = data.valid && data.reviewed;
```
mstriemer: This will poll for days if the add-on is listed. I guess you'd need to see that in the response.

I should add an `automated_signing` field. I think a reviewer can flag an unlisted add-on as needing manual review and it will never get auto-signed. In the future I guess `automated_signing` could flip from `true` to `false` if it didn't pass auto validation but the developer opted for manual review.

lib/amo-client.js
```diff
@@ -110,7 +110,12 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
+
+        if (data.processed && (failedValidation || passedValidation)) {
```
mstriemer: I find this a bit weird. Expanding `failedValidation` and `passedValidation` would simplify to `data.processed && (!data.valid || data.reviewed)` Which I think is just as clear. Maybe with a comment.

mstriemer:
We should discuss the `automated_signing` field. Sorry this is so complicated. This is sort of why I wanted there to be two requests initially but I think even with this craziness this is simpler than two requests.

lib/amo-client.js
kumar303: ah, thanks

lib/amo-client.js
kumar303: ugh, I think we'll just need to iterate and revisit this. Currently there is a timeout on polling so maybe I'll give that exception a little more info about the current state.

lib/amo-client.js
kumar303: so it's safe to assume `reviewed==true` implies `valid==true` ?

mstriemer: It would need to be valid to evaluate the right hand of the or so it isn't needed in that expression. There could be a point in the future when valid is `false` and it is reviewed (we currently support submissions with a validation timeout on the frontend) so I wouldn't bank on that.

mstriemer:
Looks good to me. We'll just have to iterate on this as we figure out the edge cases as you mentioned.

Issue #376, opened 2015-11-05
Make authenticated API request to download signed files


kumar303:
There is a new [API endpoint](https://bugzilla.mozilla.org/show_bug.cgi?id=122175) that `jpm sign` must use to fetch signed files.

Supports https://github.com/mozilla-jetpack/jpm/issues/327

PR #377, opened 2015-11-05
Fix downloading of signed files

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/376

### DIFF
--- a/test/unit/test.amo-client.js
+++ b/test/unit/test.amo-client.js
@@ -17,10 +17,11 @@ describe('amoClient.Client', function() {
 
   function setUp() {
     var self = this;
+    this.apiUrlPrefix = 'http://not-a-real-amo-api.com/api/v3';
     this.client = new amoClient.Client({
       apiKey: 'fake-api-key',
       apiSecret: 'fake-api-secret',
-      apiUrlPrefix: 'http://not-a-real-amo-api.com/api/v3',
+      apiUrlPrefix: this.apiUrlPrefix,
       signedStatusCheckInterval: 0,
       fs: {
         createReadStream: function() {
@@ -284,7 +285,7 @@ describe('amoClient.Client', function() {
         expect(result.success).to.be.equal(true);
         expect(createWriteStream.call[0]).to.be.equal(
           path.join(process.cwd(), 'some-signed-file-1.2.3.xpi'));
-        expect(fakeRequest.call[0].uri).to.be.equal(files[0].download_url);
+        expect(fakeRequest.call[0].url).to.be.equal(files[0].download_url);
         done();
       }).catch(done);
     });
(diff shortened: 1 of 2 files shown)


test/unit/test.amo-client.js
```diff
@@ -328,18 +328,57 @@ describe('amoClient.Client', function() {
...
+      expect(conf.headers).to.have.keys(['Accept', 'Authorization']);
+    });
+
+    it('preserves request headers', function() {
+      var headers = {'X-Custom': 'thing'};
+      var conf = this.client.configureRequest({
+        url: '/path',
+        headers: headers,
+      });
+      expect(conf.headers['X-Custom']).to.be.equal(headers['X-Custom']);
```
mstriemer: I'd rather see: `to.be.equal('thing')`

test/unit/test.amo-client.js
```diff
@@ -328,18 +328,57 @@ describe('amoClient.Client', function() {
...
+    });
+
+    it('allows you to override request headers', function() {
+      var headers = {'Accept': 'text/html'};
+      var conf = this.client.configureRequest({
+        url: '/path',
+        headers: headers,
+      });
+      expect(conf.headers['Accept']).to.be.equal(headers['Accept']);
```
mstriemer: Same here. Wouldn't be too hard for some typos to make this `undefined === undefined`

test/unit/test.amo-client.js
```diff
@@ -328,18 +328,57 @@ describe('amoClient.Client', function() {
...
+    });
+
+    it('makes relative URLs absolute', function() {
+      var urlPath = '/somewhere';
+      var conf = this.client.configureRequest({url: urlPath});
+      expect(conf.url).to.be.equal(this.client.apiUrlPrefix + urlPath);
```
mstriemer: This should either check the value of `this.client.apiUrlPrefix` or set it in the test.

mstriemer:
r+wc

test/unit/test.amo-client.js
kumar303: why? then `thing` would have to be updated in two places

kumar303: just saw your rationale about `undefined`. Ugh. Good point.

mstriemer: You're never going to update it so it doesn't matter. And find/replace is going to sort that out in a matter of seconds.

It's more clear. When someone looks at this they don't have to look back to verify that `headers['X-Custom']` actually has a meaningful value and there's no chance of typoing the key and having the test stop working.

test/unit/test.amo-client.js
kumar303: I moved it to a test instance var which should add some protection against `undefined` https://github.com/kumar303/jpm/blob/fix-download/test/unit/test.amo-client.js#L20

PR #380, opened 2015-11-06
Don't show additional firstrun page on official builds

gijsk:
This was pointed out on IRC, see http://logs.glob.uno/?c=mozilla%23jetpack&s=6+Nov+2015&e=6+Nov+2015#c124782 and other  bits of that log.

This page was added for official branding (release + beta) in https://bugzilla.mozilla.org/show_bug.cgi?id=1209140 . We should override it so people testing their add-on don't see it every. single. time.

@jsantell , can you take a look?


jsantell:
Thanks @gijsk !

Issue #374, opened 2015-11-04
Fix the tests


andymckay:
We've got a pile of pull requests, but the tests are all failing for them. To keep the awesome pull requests coming, I think first we should fix the tests, then we can look at the pull requests and hopefully run them against the pull requests again.

johannhof:
+1000

I thought #333 fixed tests though, maybe just give'em a restart on Travis?

johannhof:
While at it, we should also update our Travis config to some less ancient version of Node.

andymckay:
If that's the case I think the pull-requestee can re base their branch on #333 and re-kick the tests. Would be cool to merge these PRs, but so many failing tests on them.

rpl:
@andymckay @johannhof unfortunately in #333 we fixed all the tests which fail consistently, but there are still a couple of intermittent test cases (they seem to fail for timeout reasons).

In #333 we decided to merge and then check how frequent they are, but now it is evident that they are pretty frequent so I'm going to try to get the time to give it another (deeper) look.

rpl:
@andymckay @johannhof @kumar303 
During this weekend I fought with these intermittents tests, and I have two sad news and one good news:

- sad news: I'm unable to reproduce the problem locally, but it happens pretty often on travis,
  I tried different strategies and tried to run the same job on travis as much times as needed to make it fails  (and because of this travis probably hates me a lot now :-))

- good news: in #384 introduce a small change on one of the greatest test case offender: "jpm run addons - chrome", which presented the higher frequency of timeouts

- sad news: a jpm process which timeouts on a test case can (and it does) generate a random failure in another test case (because it runs "https://github.com/mozilla-jetpack/jpm/blob/90632d4/test/utils/index.js#L102" when another test case is running), I'm trying different strategies here but none that I like and which is effective

johannhof:
First of all thanks for the effort :+1:

I can't put much time into it right now, but is there any way we can approach these tests from a different angle, e.g. completely rewrite they way they work instead of hacking around all these intermittents? That might even be faster now.

erikvold:
+1

rpl:
@johannhof unfortunately (imho) rewriting the test suite is not an option until we have a precise understanding on what generates the intermittent timeout failures (or we can end up with the same issue on the rewrote test suite).

Some of the test cases that turn into timeout failures during my experimentations were actually completely successfully but the Firefox instance which runs the test addon did not exit after complete the code they had to run without an evident reason.

[...]

PR #384, opened 2015-11-09
Fix test run chrome addon intermittent failure

rpl:
This PR reduces the number of intermittent failures (and help with #374) by tweaking the test case *"jpm run addons - chrome"*:
- retrying panel.show until it succeed.

### DIFF
--- a/test/addon_runners/chrome/main.js
+++ b/test/addon_runners/chrome/main.js
@@ -68,13 +68,19 @@ exports.testChromeLocale = function(assert) {
                'locales en-US folder was copied correctly');
 }
 
+var { setInterval, clearInterval } = require("sdk/timers");
+
 exports.testChromeInPanel = function(assert, done) {
   let panel = Panel({
     contentURL: 'chrome://test/content/panel.html',
     contentScriptWhen: 'start',
     contentScriptFile: data.url('panel.js')
   });
+
+  var intervalId;
+
   panel.once('show', _ => {
+    clearInterval(intervalId);
     assert.pass('panel shown');
     panel.port.once('echo', _ => {
       assert.pass('got echo');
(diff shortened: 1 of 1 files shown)


test/addon_runners/chrome/main.js
```diff
@@ -84,7 +90,13 @@ exports.testChromeInPanel = function(assert, done) {
     });
     panel.port.emit('echo');
   });
-  panel.show();
+
+  // NOTE: hack needed to prevents random test timeout (probably due to the
+  // setTimeout used in the 'panel.show' method:
+  // https://github.com/mozilla/addon-sdk/blob/a19bd8c/lib/sdk/panel/utils.js#L78)
+  intervalId = setInterval( _ => {
```
kumar303: looks like a good enough approach! :) My only worry is that if `show()` is never called then this interval would live on even if the test times out. Or will it die along with the timeout? If there's no way to address that then no worries, but here are some ideas:

* is there a `afterEach()` or `tearDown()` hook that can be used here to clear the interval?
* should the interval give up after several tries?

kumar303:
My comments are more food-for-thought and could be addressed in follow-up commits. This looks like a decent approach to addressing the timeouts so I'll go ahead and merge it as is. Thanks!

test/addon_runners/chrome/main.js
rpl: @kumar303 this test case actually runs inside the Firefox instance launched by the jpm's mocha test case and it is not reused for running the other test addons in the jpm test suite.

In any case it is absolutely reasonable to cleanup the setInterval after the test timeouts.

I'm going to add it in a follow-ups.

rpl:
@kumar303 I agree absolutely, there are  a couple of more details that I would like to look into to completely solve the issue.

And in the mean time we will be able to see if the Travis builds are more stable on building the next pull requests

kumar303:
@rpl I've been watching the builds and they do seem more stable :) I have seen at least one other timeout, which in my haste I did not make note of. It was not in the test you addressed here though IIRC. It went away after a retry.

Issue #381, opened 2015-11-06
jpm sign should exit 1 when it fails


kumar303:
the command line program should exit 1 on failure so that developers can integrate it with other tools

Supports https://github.com/mozilla-jetpack/jpm/issues/327

PR #385, opened 2015-11-09
Make `jpm sign` exit 1 or 0 as expected

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/381

### DIFF
--- a/lib/sign.js
+++ b/lib/sign.js
@@ -3,16 +3,20 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
 
+var _ = require("lodash");
 var getID = require("jetpack-id");
 
-var AMOClient = require("./amo-client").Client;
+var DefaultAMOClient = require("./amo-client").Client;
 var xpi = require("./xpi");
 
 
-function sign (manifest, program, options) {
-  var postURL = options.postUrl;
+function sign (manifest, program, options, config) {
+  config = _.assign({
+    systemProcess: process,
+    AMOClient: DefaultAMOClient,
+  }, config);
 
-  var client = new AMOClient({
+  var client = new config.AMOClient({
     apiKey: options.apiKey, apiSecret: options.apiSecret,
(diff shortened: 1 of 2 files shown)


lib/sign.js
```diff
@@ -26,8 +31,11 @@ function sign (manifest, program, options) {
 
   }).then(function(result) {
     console.log(result.success ? 'SUCCESS' : 'FAIL');
+    config.systemProcess.exit(result.sucess ? 0 : 1);
```
mstriemer: `result.sucess` ~> `result.success`

test/unit/test.sign.js
```diff
@@ -593,6 +594,92 @@ describe('amoClient.PseudoProgress', function() {
...
+    runSignCmd().then(function() {
+      expect(mockProcessExit.call[0]).to.be.equal(0);
+      done();
+    }).catch(done);
+  });
+
+  it('should exit 1 on signing failure', function(done) {
+    runSignCmd({
+      StubAMOClient: makeAMOClientStub({
+        result: {sucess: false},
```
mstriemer: type of `success` happens here too.

mstriemer:
There seems to be a lot of setup code to test one function. I'd rather see a function that just gets called directly with different values of `response.success`. Since you added the `describe('sign', ...)` block then maybe this setup code will eventually be required but it looks out of place to what is currently being tested.

r+wc

test/unit/test.sign.js
kumar303: d'oh! auto-completion fail

kumar303: huh, I tested it on the command line too (maybe I only checked errors).

kumar303:
> There seems to be a lot of setup code to test one function. 

Agreed. It's a high level function that integrates all other components to run the complete shell command. These are always tricky to test. I made it a little more like the `xpi` command test which let me remove at least one of the stubs.

Issue #387, opened 2015-11-10
jpm sign: show a friendly error when key/secret is not specified


kumar303:
If `--api-key` or `--api-secret` is empty, you get a stack trace which might be confusing. 

Supports https://github.com/mozilla-jetpack/jpm/issues/327

PR #388, opened 2015-11-11
Make key/secret required options for jpm sign

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/387

Looks like:
````
$ jpm sign
JPM [info] Starting jpm sign on GitHub Tab Width

  error: missing required option `--api-key'
  error: missing required option `--api-secret'

FAIL
````

### DIFF
--- a/lib/sign.js
+++ b/lib/sign.js
@@ -5,6 +5,7 @@
 
 var _ = require("lodash");
 var getID = require("jetpack-id");
+var when = require("when");
 
 var DefaultAMOClient = require("./amo-client").Client;
 var xpi = require("./xpi");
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/sign.js [hunk 1]
```diff
@@ -16,14 +17,36 @@ function sign (manifest, program, options, config) {
...
-    apiUrlPrefix: options.apiUrlPrefix,
-  });
+  return when.promise(function(resolve, reject) {
+
+    var missingOptions = [];
+    [
+      {value: options.apiKey, flag: '--api-key'},
+      {value: options.apiSecret, flag: '--api-secret'},
+    ]
+    .forEach(function(opt) {
```
mstriemer: I think this looks really weird. I'd rather see the `forEach` on the same line as the end of the array or giving the array a name. At first glance I thought this was supposed to be assigned to `missingOptions`.

I also find it strange that commander won't handle this. Oh well.

mstriemer:
r+wc

lib/sign.js
kumar303: Yeah, I did hunt through the docs and source and I don't think commander handles it. Admittedly, required "options" is a bit odd but I don't think defining these as arguments is any nicer because you lose context like `key=...` and `secret=...`

PR #390, opened 2015-11-12
Adding failed exits for errors on jpm test

drazisil:
Hopefully I placed these exits in the correct place, it's kinda hard to follow the flow.

Reason for change: CI systems that check for return code on tests will pass failures, these changes seek to cause them to fail.

### DIFF
--- a/lib/firefox.js
+++ b/lib/firefox.js
@@ -59,6 +59,7 @@ function runFirefox (manifest, options) {
         console.error("No Firefox binary found at " + binary);
         if (!options.binary) {
           console.error("Specify a Firefox binary to use with the `-b` flag.");
+          process.exit(1)
         }
       }
       else {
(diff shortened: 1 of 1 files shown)


drazisil:
@kumar303, @gozala, @jsantell, @mossop Can I get a review/feedback please? :)

kumar303:
Hi. Thanks for taking the time to submit a patch.

Instead, I think it would be better to call `process.exit(1)` at the very end of the promise chain which, for `jpm test`, is in [lib/test](https://github.com/drazisil/jpm/blob/patch-1/lib/test.js#L10). For reference, the `jpm sign` command already [has support for this](https://github.com/drazisil/jpm/blob/patch-1/lib/sign.js#L37). Otherwise, with your patch, the exit status will only be set for certain error conditions. We'd want exit status set for all errors.

I know the test suite is heavy on functional tests but this patch should get a test added for it too. I think a unit test using a mock `process` object would be fine, which is how `jpm sign` [is tested](https://github.com/drazisil/jpm/blob/patch-1/test/unit/test.sign.js#L674).

kumar303:
I'm closing this for lack of feedback but feel free to re-open or submit a new patch.

PR #391, opened 2015-11-12
Consider the xpi signed only when files are ready

kumar303:
I have seen this occur both locally and on stage where the file was processed and had validation results but the files array was empty. When inspecting the DB, signed files were there so I suspect it's a race condition. It may be better fixed by adding a db transaction somewhere but, meh, that code is easy to get lost in ;) I couldn't find an obvious place to patch it.


mstriemer:
r+

PR #394, opened 2015-11-17
Run jpm sign on a custom XPI file

### LINKED ISSUE #393: jpm sign: pass in any XPI filename for signing
This would add a feature to jpm sign where it can sign any XPI file, not just the SDK-based one in your current working directory. This would allow non-SDK add-on devs to use `jpm sign`. Eventually we want to provide a standalone command for them but we're currently bikeshedding the name of it! So it might be a while ;)

kumar303:
Fixes #393


lib/cmd.js
```diff
@@ -47,3 +31,22 @@ function isEmptyCommand (program) {
   return false;
 }
 exports.isEmptyCommand = isEmptyCommand;
+
+function validateProgram(program) {
+  Object.keys(program).filter(function (option) {
+    return ~DEBUG_VALUES.indexOf(option);
```
mstriemer: So it seems as though this will always work but I'd rather this check if the result is `< 0`.

I know you didn't write it but I have a feeling other people will look at this and go "wtf?" then start trying it out and eventually realise that two's complement for bitwise operators is in the spec and so is `indexOf` returning `-1` so it *should* always work but really do we want to count on that? And why use `~` when you could just compare to `-1`? Oh well.

lib/sign.js
```diff
@@ -37,25 +40,41 @@ function sign (manifest, program, options, config) {
...
+      var xpiConfig = {
+        manifest: manifest,
+      };
+      if (options.xpi) {
+        logger.log("Signing XPI: " + options.xpi);
+        return _.assign(xpiConfig, {xpiPath: options.xpi});
+      } else {
+        return xpi(manifest, options).then(function(xpiPath) {
+          logger.log("Created XPI for signing: " + xpiPath);
+          return _.assign(xpiConfig, {xpiPath: xpiPath});
```
mstriemer: Seems like you could just call it `path` if you wanted since it's already about an XPI.

lib/utils.js
```diff
@@ -75,19 +78,60 @@ exports.console = Object.freeze(jpmConsole);
...
+    if (xpiPath) {
+      return resolve(getXPIManifest(xpiPath));
+    } else {
+      var cwd = process.cwd();
+      var json = path.join(cwd, "package.json");
+      var manifest = {};
+      try {
+        manifest = require(json);
+      }
+      catch (e) {}
```
mstriemer: Looks like the `catch` can come after the `}`. I see both forms in the codebase but this file has a `} catch(e) {` in it so this should probably be consistent with that (and I'm glad that this file has the other one to support my case :wink:).

lib/utils.js
```diff
@@ -75,19 +78,60 @@ exports.console = Object.freeze(jpmConsole);
...
-  })
+  });
 }
 exports.getManifest = getManifest;
+
+
+function getXPIManifest(xpiPath) {
+  return when.promise(function(resolve, reject) {
+    tmp.dir(function(err, tmpPath, removeTmpDir) {
+      if (err) {
```
mstriemer: Looks like the temporary directory isn't always being Removed. Seems like you might want to make a new reject function to use instead that does something like:

```js
function rejectWithCleanup(err) {
  removeTmpDir();
  reject(err);
}

File: lib/utils.js [hunk 5]
```diff
@@ -75,19 +78,60 @@ exports.console = Object.freeze(jpmConsole);
...
+
+    }, {
+      // This allows us to remove a non-empty tmp dir.
+      unsafeCleanup: true,
+    });
+  })
+  .then(function(tmpDir) {
+    var manifest = {};
+    try {
+      manifest = require(path.join(tmpDir.path, "package.json"));
```
mstriemer: Couldn't this move into the `close` handler and have the cleanup happen there? Doesn't really matter I guess but I was expecting the cleanup to be in there.

test/unit/test.sign.js
```diff
@@ -648,17 +649,69 @@ describe('sign', function() {
...
+  });
+
+  it('passes custom XPI to the signer', function(done) {
+    var mockManifestGetter = new CallableMock({
+      returnValue: when.promise(function(resolve) {
+        resolve({});  // resolve with an empty manifest.
+      }),
+    });
+    // Make sure nothing is checking the working directory for add-on things.
+    process.chdir(path.join(__dirname, "..", "addons"));
```
mstriemer: I don't see it switching back. Is this safe? Can you put a real XPI on the filesystem and skip the mock?

mstriemer: Oh, you don't care here the test with the XPI is in the next file. Why is this needed then?

test/unit/test.utils.js
```diff
@@ -27,20 +28,63 @@ describe("lib/utils", function () {
...
+      process.chdir(path.join(__dirname, "..", "addons"));
+
+      utils.getManifest(simpleAddonXPI).then(function(manifest) {
+        expect(manifest.name).to.be.equal("simple-addon");
+        expect(manifest.title).to.be.equal("My Simple Addon");
+        done();
+      }).catch(done);
+    });
+
+    it("throws error for non-existant XPI file", function (done) {
```
mstriemer: These tests look good but I think you can return a promise in chai if you want instead of using `done`. Works for me either way.

```js
it("throws error for non-existant XPI file", function () {
  var badXPIFile = path.join(__dirname, "..", "not-a-real-file.xpi");
  return utils.getManifest(badXPIFile).then(function() {
    chai.fail('Unexpected success');
  }).catch(function(err) {
    expect(err.message).to.include('ENOENT');
    expect(err.message).to.include(badXPIFile);
 });
});

mstriemer:
r+wc

lib/cmd.js
kumar303: oh, I read the benchmark wrong. The `1 + ...indexOf` approach is the slowest.

Issue #370, opened 2015-11-02
jpm post shouldn't generate update.rdf


FayneAldan:

kumar303:
fixed in https://github.com/mozilla-jetpack/jpm/pull/395

PR #395, opened 2015-11-17
Should fix #369 and #370

### LINKED ISSUE #369: The Promise returned by jpm/lib/post finishes too early
`jpm/lib/post` returns a Promise that finishes when the XPI is created, rather than when the XPI is posted.

FayneAldan:
Fixes #369 and #370

### DIFF
--- a/lib/post.js
+++ b/lib/post.js
@@ -13,7 +13,7 @@ function postXPI (manifest, options) {
   var postURL = options.postUrl;
 
   return xpi(manifest, extend(options, { skipUpdateRDF: true })).then(function (xpiPath) {
-		return when.promise(function(done) {
+    return when.promise(function(resolve, reject) {
       fs.readFile(xpiPath, function(err, buffer) {
         var p = url.parse(postURL);
         var client = net.connect({ host: p.hostname, port: p.port }, function() {
@@ -29,12 +29,13 @@ function postXPI (manifest, options) {
         });
         client.on('error', function (e) {
           // if net errors need to be propagated.
+          reject(e);
         });
         client.on('close', function (had_error) {
           console.log(had_error ? "Posting XPI to %s failed" : "Posted XPI to %s", postURL);
           fs.unlink(xpiPath);
           console.log("Removed XPI from " + xpiPath);
-          done();
+          resolve();
         });
       });
     });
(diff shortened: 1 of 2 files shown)


lib/post.js
```diff
@@ -4,43 +4,38 @@
...
-      });
-      client.on('error', function (e) {
-        // if net errors need to be propagated.
-      });
-      client.on('close', function (had_error) {
-        console.log(had_error ? "Posting XPI to %s failed" : "Posted XPI to %s", postURL);
-        fs.unlink(xpiPath);
-        console.log("Removed XPI from " + xpiPath);
+  return xpi(manifest, extend(options, { skipUpdateRDF: true })).then(function (xpiPath) {
+		return when.promise(function(done) {
```
kumar303: this should only be indented two spaces past the enclosing statement

lib/post.js
```diff
@@ -4,43 +4,38 @@
...
+          client.end(buffer);
+        });
+        client.on('data', function (data) {
+          // Extension Auto-Installer does not close the connection, so we do it after 500ms
+          setTimeout(function() {
+            client.destroy();
+          }, 500);
+        });
+        client.on('error', function (e) {
+          // if net errors need to be propagated.
```
kumar303: while you're in here, you might as well change the signature to `when.promise(function(resolve, reject) {...` so you can call `reject(e)` right here. That will show a stack trace in the console and will allow the caller to handle the promise rejection.

kumar303:
I guess `post` doesn't have any tests so you're off the hook ;) This needs rebasing on master for the test suite fix though.

lib/post.js
FayneAldan: My bad. I normally use tabs when programming, instead of spaces.

lib/post.js
```diff
@@ -4,43 +4,39 @@
 "use strict";
 
 var extend = require("lodash").extend;
-var execute = require("./task");
-var http = require("http");
 var xpi = require("./xpi");
-var cp = require('child_process');
```
kumar303: thanks for cleaning these up. We should really start running a linter.

kumar303:
Looks ok to me, thanks

Issue #332, opened 2015-09-03
The XPI module shouldn't rely on process.cwd()


jsnajdr:
I'm using JPM to build several extensions at once using Gulp:

```js
var xpi = require('jpm/lib/xpi');
dirList.forEach(function(dir) {
  process.chdir(dir);
  var manifest = require(path.join(dir, 'package.json'));
  xpi(manifest, { xpiPath: destDir });
});
```

This of course doesn't work, because several async task are spawned and the process can only have one working directory. To fix it, I must build all the extensions synchronously in sequence.

It would be great if the xpi function didn't rely internally on process.cwd(), but accepted a 'srcDir' option.

jsnajdr:
Pull request merged, marking as fixed.

PR #399, opened 2015-11-18
Issue #332: Added option --addon-dir (defaults to the current work dir)

jsnajdr:
This solves #332: adds a new option to set the sources directory and override the process.cwd. This is an updated and polished commit from abandoned pull request #272

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -87,16 +87,16 @@ exports.console = Object.freeze(jpmConsole);
  *          Without this, the current working directory is assumed.
  * @return {Promise} resolves to a manifest object
  */
-function getManifest (options) {
+function getManifest(options) {
   options = _.assign({
-    xpiPath: null,
+    addonDir: null,
+    xpiPath: null
   }, options);
   return when.promise(function(resolve, reject) {
     if (options.xpiPath) {
       return resolve(getXPIManifest(options.xpiPath));
     } else {
-      var cwd = process.cwd();
-      var json = path.join(cwd, "package.json");
+      var json = path.join(options.addonDir, "package.json");
       var manifest = {};
       try {
         manifest = require(json);
--- a/test/functional/test.run.js
+++ b/test/functional/test.run.js
@@ -302,7 +302,6 @@ describe("jpm run", function () {
       });
 
       it("run with options should receive options (--filter)", function(done) {
-        process.chdir(paramDumpPath);
         var cmd = "run -v --profile-memory --check-memory --filter bar --times 3 --stop-on-error --do-not-quit --tbpl"
         var task = exec(cmd, options, function(error, stdout, stderr) {
           expect(error).to.not.be.ok;
@@ -331,7 +330,6 @@ describe("jpm run", function () {
       });
 
       it("run with options should receive options (-f)", function(done) {
-        process.chdir(paramDumpPath);
         var cmd = "run -v --profile-memory --check-memory -f bar --times 3 --stop-on-error --do-not-quit --tbpl"
         var task = exec(cmd, options, function(error, stdout, stderr) {
           expect(error).to.not.be.ok;
--- a/test/functional/test.run.js
+++ b/test/functional/test.run.js
@@ -40,7 +40,6 @@ describe("jpm run", function () {
 
   it("should not overwrite or delete xpi in current xpi", function (done) {
     var size = null;
-    process.chdir(paramDumpPath);
     // Copy over a different xpi and rename it to the would-be generated xpi name
     fs.copy(path.join(__dirname, "..", "xpis", "@aom-unsupported.xpi"), path.join(paramDumpPath, "@param-dump.xpi")).then(function () {
       return fs.stat(path.join(paramDumpPath, "@param-dump.xpi"));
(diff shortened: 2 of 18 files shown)


lib/utils.js
```diff
@@ -78,16 +78,15 @@ exports.console = Object.freeze(jpmConsole);
  *
  * @return {Object}
  */
-function getManifest () {
+function getManifest (addonDir) {
```
kumar303: Can you make this more like `getManifest ({addonDir: ...})` instead? Because [this other patch](https://github.com/mozilla-jetpack/jpm/pull/394/files#diff-50e3aa130a4f97a42ee2cf111c7b1d9dR87) needs to pass in a new option to `getManifest()` as well.

kumar303:
Can you also update the usage docs in the README ?

kumar303:
Thanks for taking the time to re-do your patch, sounds like a good feature. I can help get it merged after the fix-ups.

kumar303:
The patch from https://github.com/mozilla-jetpack/jpm/pull/394 just landed so you'll need to rebase on master for that. There should only be one conflict: I added `getManifest({xpiPath: ...})`. An easy way to do it would be to change your patch to use `getManifest ({addonDir: ...})` first, make sure the tests pass, then rebase and resolve the conflict.

jsnajdr:
Improved few things pointed out in review:
- arguments to getManifest are an options object with properties - compatible with the signing patch
- updated README.md documentation
- put back pushd/popd behavior into test before/after handlers
- fixed the sign command to work well with the addonDir option
- cleaned up the tests - return promise to Mocha instead of calling done()

One remaining issue:
- I could not push README.md that contained string "--addon-dir". The spell checking test detected it as spelling error. So I had to change it to "--dir". Should be fixed by someone more competent than me before merging.

test/functional/test.run.js
```diff
@@ -288,7 +276,6 @@ describe("jpm run", function () {
       }
 
       it("run with only -v option", function(done) {
-        process.chdir(paramDumpPath);
```
kumar303: after this, does the cmd need `{ addonDir: paramDumpPath ...` ? I can't quite tell what it's testing. I also noticed some `chdir(paramDumpPath)` calls are still in tests right after this but that's probably fine.

test/unit/test.sign.js
```diff
@@ -599,16 +599,13 @@ describe('amoClient.PseudoProgress', function() {
 
 describe('sign', function() {
   var simpleAddonPath = path.join(__dirname, "..", "fixtures", "simple-addon");
-  var manifest;
```
kumar303: whoops, good catch

test/unit/test.utils.js
```diff
@@ -29,18 +29,14 @@ describe("lib/utils", function () {
...
+    it("returns manifest in simple-addon directory", function () {
+      return utils.getManifest({addonDir: simpleAddonPath}).then(function(manifest) {
         expect(manifest.name).to.be.equal("simple-addon");
         expect(manifest.title).to.be.equal("My Simple Addon");
       });
     });
 
     it("returns manifest from custom XPI file", function () {
-      // Change into a non-package path.
-      process.chdir(path.join(__dirname, "..", "addons"));
```
kumar303: This is a bit weird but what it did was made sure the test uses `simpleAddonXPI` rather than the add-on dir. So I think you should make this additional change :
````
return utils.getManifest({
    xpiPath: simpleAddonXPI, 
    addonDir: path.join(__dirname, "..", "addons"),  // pass in an invalid dir just to make sure it's not used
  })
    ...
````

File: test/utils/index.js [hunk 5]
```diff
@@ -54,7 +54,8 @@ function exec (args, options, callback) {
   options = options || {};
   var env = _.extend({}, options.env, process.env);
 
-  return child_process.exec("node " + jpm + " " + args, {
+  return child_process.exec("node " + jpm + " --addon-dir " + options.addonDir
```
kumar303: what happens if `options.addonDir` is undefined? That could potentially break a test no? I think it would be a little safer as:
````
if (options.addonDir) {
  args = "--addon-dir " + options.addonDir + " " + args;
}
return child_process.exec("node " + jpm + " " + args ...
````

test/functional/test.run.js
jsnajdr: The other chdir's that are remaining in this file are useless and I'll remove them. Using the addonDir option is the right way. The tests are running the 'params-dump' addon in the fixtures directory and are testing its output.

kumar303:
Everything looks good! Thanks for the rapid follow-ups.

I'm on vacation next week but when I'm back (Nov 30) I'd like to check out your branch locally and just test it out a bit since it covers pretty much all jpm commands. Then I'll merge it in.

There were a couple signing bug fixes that just landed but nothing that will create conflicts. However, if you can rebase your changes on master then that would be helpful when I manually test it out.

kumar303:
Thanks for your patience. I just tested it out locally and merged it into master. I only found one issue which was an easy fix: https://github.com/mozilla-jetpack/jpm/commit/af0859c5bc7585dbbec65a45866bb90d1cf5b824 Thanks again for the patch and spending so much time to get it in.

jsnajdr:
Thanks for the merge! Always happy to contribute to Firefox :)

danilvalov:
The package hasn't been updated in npm :(

kumar303:
@danilvalov I just pushed a release https://www.npmjs.com/package/jpm Sorry about the delay.

danilvalov:
@kumar303 Thank you!

PR #402, opened 2015-11-20
Listen to write-stream end for downloads

kumar303:
For some reason the old way of listening to 'end' on the response was saving incomplete files fetched from the CDN. This approach works.


mstriemer:
It's not just the follow redirect that fixes it is it? Looks like you just moved the on calls into the object rather than the return value. If it works it works though. r+

kumar303:
> It's not just the follow redirect that fixes it is it?

There was no change here. The default value for this option is the same. I just thought it might be a good idea to set it explicitly since we rely on it.

> Looks like you just moved the on calls into the object rather than the return value. If it works it works though

Yeah, same pipe-able interface. It seemed a little easier to read this way.

I tested on stage and prod -- Nightly no longer complains about a corrupt XPI file.

Issue #386, opened 2015-11-10
jpm sign: poll for status using an API-provided URL


kumar303:
Now that we have a direct URL to check the status of an upload (https://github.com/mozilla/olympia/pull/896), `jpm sign` should use that URL to check status instead of the addon + version URL it was using before. This will make signing more reliable if concurrent runs of `jpm sign` are in progress.

Supports https://github.com/mozilla-jetpack/jpm/issues/327

PR #403, opened 2015-11-20
Poll signing status with exact upload status URL

kumar303:
This avoids concurrency mishaps in case multiple uploads are in the upload queue.

Fixes https://github.com/mozilla-jetpack/jpm/issues/386


mstriemer:
r+

PR #410, opened 2015-12-02
Handle foreign zipped XPIs better

kumar303:
This addresses bugs in the node unzip module which caused 'Error: invalid signature' when parsing certain XPIs that were not created by jpm.


johannhof:
This could also be fixed by #521, please reopen if I'm mistaken.

Issue #412, opened 2015-12-02
jpm sign sometimes says your unlisted add-on is waiting for review 


kumar303:
The patch in https://github.com/mozilla-jetpack/jpm/pull/408 introduced a problem because sometimes a processed result is `active: false` before the signed files have been added. This is a timing issue because multiple celery tasks are involved in the signing and there is a delay until all are complete. Sigh.

kumar303:
We're waiting on this API change to be deployed to production: https://github.com/mozilla/olympia/pull/1041 ETA: Dec 10, 2015

kumar303:
Fixed in https://github.com/mozilla-jetpack/jpm/pull/415

Issue #407, opened 2015-11-30
jpm sign times out for listed + full review add-ons


kumar303:
To reproduce:

- submit an add-on to AMO and make it listed and request a full review
- get the initial add-on approved
- create a new version
- run `jpm sign` to auto-update the new version

This always ends in a time-out, with the following output:

````
$ jpm sign --api-key ... --api-secret ... --api-url-prefix http://olympia.dev/api/v3 --xpi ~/tmp/addons/scratch1/\@scratch1-0.0.4.xpi
JPM [info] Signing XPI: /Users/kumar/tmp/addons/scratch1/@scratch1-0.0.4.xpi
Validating add-on [......................................................................................................................................................................................]
JPM [error] FAIL
Error: Validation took too long to complete; last status: {"active":false,"url":"http://olympia.dev/api/v3/addons/%40scratch1/versions/0.0.4/uploads/256ab54aeb8f4163a264c57fb32848e4/","files":[{"hash":"sha256:6aab8ba9f66388e0e804d947c66f9a28ec8e0f595612a774ec6858959585640e","signed":false,"download_url":"http://olympia.dev/api/v3/file/132/scratch_1-0.0.4-fx+an.xpi?src=api"}],"passed_review":false,"pk":"256ab54aeb8f4163a264c57fb32848e4","processed":true,"reviewed":false,"valid":true,"validation_results":{"errors":0,"success":true,"warnings":0,"compatibi...
    at null._onTimeout (/Users/kumar/dev/jpm/lib/amo-client.js:151:14)
    at Timer.listOnTimeout (timers.js:110:15)
````

This is because the file will always require manual review since listed + full review gives it a side-loading certificate

kumar303:
I think this would also be reproducible by submitting an unlisted add-on and checking the box for 'this will be bundled with an installer'

kumar303:
This wasn't the right solution because it broke other things. See https://github.com/mozilla-jetpack/jpm/pull/413

kumar303:
When the API changes from https://github.com/mozilla/olympia/pull/1041 are live (ETA: Dec 10, 2015), this patch should resolve the issue https://github.com/mozilla-jetpack/jpm/pull/415 I'll leave this open until then.

kumar303:
This went out on Dec 10, 2015, as you can see from the tag: https://github.com/mozilla/olympia/commit/9b4d02927e078aa68fc343b356a2d517e2a7b5f1

[...]

PR #413, opened 2015-12-03
Removes buggy detection of listed signing

kumar303:
This addresses https://github.com/mozilla-jetpack/jpm/issues/412 but will revert https://github.com/mozilla-jetpack/jpm/issues/407 temporarily because I think we need to wait on an API change.


kumar303:
This can be reverted once the API changes from https://github.com/mozilla/olympia/pull/1041 have been deployed to production.

kumar303:
I have a better idea for how to deal with the workaround: https://github.com/mozilla-jetpack/jpm/pull/415

Issue #414, opened 2015-12-03
jpm sign: handle results for listed add-ons better


kumar303:
We need to look for a new API field added in https://github.com/mozilla/olympia/pull/1041 to know how to properly wait on listed add-on signing results without timing out (https://github.com/mozilla-jetpack/jpm/issues/407).

Current workaround which should be removed: https://github.com/mozilla-jetpack/jpm/pull/413

PR #415, opened 2015-12-03
Detect signing results for listed add-ons better

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/414 and will eventually address https://github.com/mozilla-jetpack/jpm/issues/407 (when API changes are deployed).

### DIFF
--- a/lib/amo-client.js
+++ b/lib/amo-client.js
@@ -121,14 +121,24 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
         var data = result[1];
         lastStatusResponse = data;
 
+        // TODO: remove this when the API has been fully deployed with this
+        // change: https://github.com/mozilla/olympia/pull/1041
+        var apiReportsAutoSigning = typeof data.automated_signing !== 'undefined';
+
+        var canBeAutoSigned = data.automated_signing;
         var failedValidation = !data.valid;
         // The add-on passed validation and all files have been created.
-        var signedAndReady = (data.valid && data.active && data.reviewed &&
-                              data.files && data.files.length > 0);
-        // The add-on is valid but requires a manual review before it can be signed.
-        var requiresManualReview = data.valid && !data.active && !data.reviewed;
+        // There are many checks for this state because the data will be
+        // updated incrementally by the API server.
+        var signedAndReady = data.valid && data.active && data.reviewed &&
+                             data.files && data.files.length > 0;
+        // The add-on is valid but requires a manual review before it can
+        // be signed.
+        var requiresManualReview = data.valid && apiReportsAutoSigning &&
+                                   !canBeAutoSigned;
 
-        if (data.processed && (failedValidation || signedAndReady || requiresManualReview)) {
+        if (data.processed &&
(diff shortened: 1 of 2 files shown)


lib/amo-client.js
```diff
@@ -121,14 +121,24 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
         // The add-on passed validation and all files have been created.
+        // There are many checks for this state because the data will be
+        // updated incrementally by the API server.
         var signedAndReady = (data.valid && data.active && data.reviewed &&
                               data.files && data.files.length > 0);
-        // The add-on is valid but requires a manual review before it can be signed.
-        var requiresManualReview = data.valid && !data.active && !data.reviewed;
+        // The add-on is valid but requires a manual review before it can
+        // be signed.
+        var requiresManualReview = (data.valid && apiReportsAutoSigning &&
```
mstriemer: JavaScript doesn't actually require these parens. Same on line 133. I'm fine with them though if you prefer it.

lib/amo-client.js
```diff
@@ -121,14 +121,24 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
+                                    !canBeAutoSigned);
 
-        if (data.processed && (failedValidation || signedAndReady || requiresManualReview)) {
+        if (data.processed &&
+              (failedValidation || signedAndReady || requiresManualReview)) {
```
mstriemer: I can be fail validation and be signed now.

mstriemer: Wait no, nevermind.

lib/amo-client.js
```diff
@@ -121,14 +121,24 @@ Client.prototype.waitForSignedAddon = function(statusUrl, opt) {
...
+
+        var canBeAutoSigned = data.automated_signing;
         var failedValidation = !data.valid;
         // The add-on passed validation and all files have been created.
+        // There are many checks for this state because the data will be
+        // updated incrementally by the API server.
         var signedAndReady = (data.valid && data.active && data.reviewed &&
                               data.files && data.files.length > 0);
-        // The add-on is valid but requires a manual review before it can be signed.
-        var requiresManualReview = data.valid && !data.active && !data.reviewed;
```
mstriemer: The original bug was that this fallback doesn't work, so it has to be removed, right?

kumar303: Yeah, this doesn't work because a result for an unlisted add-on might satisfy these constraints for a couple seconds -- right before the actual files have been signed and the version table has been updated with the file objects. I was seeing it pretty often. If I used curl a couple seconds later to check the status URL I could see that the files were in there.

lib/amo-client.js
kumar303: oh yeah, heh. I thought we had a lint rule against it in other repos but maybe not. This repo doesn't have lint anyway.

mstriemer:
I think that's right. I don't find the extra `var`s very helpful so I would be cool with inlining most of those boolean expressions but it's fine with them too.

r+

kumar303:
Just tested this on dev and it's working. On stage (without the new API data) it times out for listed add-ons as expected.

PR #416, opened 2015-12-04
Fix message about valid, listed add-ons

kumar303:
This is due to a policy change: https://blog.mozilla.org/addons/2015/12/01/de-coupling-reviews-from-signing-unlisted-add-ons/


andymckay:
r+

Issue #398, opened 2015-11-18
jpm sign: clean up the temporarily created XPI file


kumar303:
After signing, the generated XPI file (not the signed one) can be removed.

Supports https://github.com/mozilla-jetpack/jpm/issues/327

tofutim:
It would also be nice to specify a bld directory.

PR #417, opened 2015-12-04
Remove auto-generated, unsigned XPI after signing

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/398

### DIFF
--- a/lib/sign.js
+++ b/lib/sign.js
@@ -4,6 +4,7 @@
 "use strict";
 
 var _ = require("lodash");
+var fs = require("fs");
 var getID = require("jetpack-id");
 var when = require("when");
 
@@ -20,6 +21,7 @@ function sign (program, options, config) {
     getManifest: utils.getManifest,
     systemProcess: process,
     AMOClient: DefaultAMOClient,
+    removeFile: fs.unlinkSync,
   }, config);
 
   return when.promise(function(resolve, reject) {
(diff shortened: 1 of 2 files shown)


test/unit/test.sign.js
```diff
@@ -894,6 +898,48 @@ describe('sign', function() {
...
+      getManifest: mockManifestGetter.getCallable(),
+      createXPI: mockXPICreator.getCallable(),
+    }).then(function() {
+      expect(fakeFileRemover.call[0]).to.be.equal(unsignedXpiPath);
+      expect(mockProcessExit.call[0]).to.be.equal(0);
+      done();
+    }).catch(done);
+  });
+
+  it('does not remove custom XPI after signing', function(done) {
```
mstriemer: I feel like this should state why the XPI is not removed.

test/unit/test.sign.js
```diff
@@ -894,6 +898,48 @@ describe('sign', function() {
     }).catch(done);
   });
 
+  it('removes created XPI after signing', function(done) {
```
mstriemer: And why is it removed?

mstriemer:
r+wc

lib/sign.js
```diff
@@ -50,13 +52,13 @@ function sign (program, options, config) {
         };
         if (options.xpi) {
           logger.log("Signing XPI: " + options.xpi);
-          return _.assign(xpiConfig, {xpiPath: options.xpi});
+          return _.assign(xpiConfig, {xpiPath: options.xpi, removeUnsigned: false});
         } else {
           // Merge the program and sub-command options.
           var xpiOptions = _.assign({}, program, options);
           return config.createXPI(manifest, xpiOptions).then(function(xpiPath) {
```
mstriemer: Actually, can you get `config.createXPI` to make the XPI in a temporary directory? I'm guessing it doesn't support that but if it does that'd be cool.

test/unit/test.sign.js
kumar303: The jpm sign command isn't responsible for creating unsigned files. You can use `jpm xpi` for that. I don't think this ever should have kept the XPI around. All the other commands that create temporary XPIs [remove them](https://github.com/mozilla-jetpack/jpm/blob/master/lib/task.js#L45) afterwards. I didn't realize that because I was mainly modeling `jpm xpi`.

test/unit/test.sign.js
kumar303: because it's a custom XPI, which is part of the test name. I can change it to external.

lib/sign.js
kumar303: oh, good idea. Yeah, it looks like there are already options I can use for that

PR #419, opened 2015-12-07
Localization for description and title at manifest file, set xpi file name at coomand line

matraska23:
Hi everybody!

I'm developing a one plugin for Firefox browser, and i was needed at some features for automatically add on building.
So i add them at my jpm fork...

1) Set xpi file name with option
```
#: jpm xpi --out <filename>
```
2) Add localization for description and title at manifest file
I know there exist a similar bug (https://bugzilla.mozilla.org/show_bug.cgi?id=436344). But i think, it's not correct, because jetpack don't do anything with language preference files.
So i add localized variants of description at manifest file.
```
"localized_description": {
 "ru": {
    "title": "Addon tittle on another language",
    "description": "Addon description on another language"
 }
},
```
3) Add page for addon configuration at manifest

Set link on addon settings page:
```
"optionsURL": "data/page/option_page.html",
```


matraska23:
I know that threr are three features at one pull request... but there are usefull.

README.md
```diff
@@ -52,6 +53,22 @@ npm link
   every time a file in the current working directory changes.
 * `jpm sign` Retrieve a Mozilla-signed `.xpi` file for your current add-on.
 
+### Fork features
```
kumar303: Could you provide a real heading?

README.md
```diff
@@ -40,6 +40,7 @@ npm link
 * `-o, --overload <path>` Uses either the specified `<path>` or the path set in the environment variables `JETPACK_ROOT` as the root for [addon-sdk](https://github.com/mozilla/addon-sdk) modules instead of the ones built into Firefox.
 * `--addon-dir <path>` Specify a source directory of the add-on (instead of current directory) when building an add-on with `xpi`.
 * `--post-url <URL>` **experimental** Used to post a xpi to a URL.
+* `--out <filename>` Override default xpi file name. Work with `jpm xpi`
```
kumar303: These sentences read awkwardly. Can you reword it?

kumar303:
Hi. Thanks for taking the time to submit a patch. Since these introduce several new features that we wouldn't want to break later on, they need unit tests. Let us know if you have need ideas for how to write those tests.

@rpl could you review the RDF and localization changes? I don't know understand enough about SDK bootstrapping to know what's going on here.

README.md
matraska23: Yes, i can.

README.md
matraska23: I change the heading.
P.S. best place for this section is  addon manifest documentation

matraska23:
I can't write correct test, because i made copy of one (from /test folder) and it fall down with exception `Uncaught AssertionError: expected false to equal true`.
Stack trace begin from: `at ChildProcess.<anonymous> (<...>\jpm\test\utils\index.js:109:29)`
And i haven't ideas how to fix it

kumar303:
About the tests: ok, I understand. You have run into a known issue.

rpl:
@kumar303 @matraska23 I think the best approach for this pull request is to focus on the new command line option.

jpm has already support for advanced install.rdf customization: 
if there is an install.rdf file in the add-on source dir, then it will be included in the generated xpi
(https://github.com/mozilla-jetpack/jpm/blob/master/lib/xpi/utils.js#L53)

for simpler (and most common) scenarios:
- in the Addon SDK the "simple-prefs" module (which generates the option page from the add-on preferences defined in the package.json)
- name and description localization can be done on AMO (addons.mozilla.org)

bin/jpm
```diff
@@ -38,7 +38,8 @@ program
   .option("--debug", "Enable the add-on debugger when running the add-on")
   .option("--check-memory", "Enable leaked tracker that attempts to report compartments leaked")
   .option("--profile-memory", "Enable profiling of memory usage")
-  .option("--retro", "In development flag for transitioning to new style addons; forces the lack of install.rdf/bootstrap.js creation regardless of what engine versions are running");
+  .option("--retro", "In development flag for transitioning to new style addons; forces the lack of install.rdf/bootstrap.js creation regardless of what engine versions are running")
+  .option("--out <fname>", "Save xpi file with this name. Use with `jpm xpi`.");
```
rpl: @kumar303 @matraska23 I would prefer an option name which is more specific than "--out", even better if it doesn't start with an 'o' (which is already used in the "-o, --overloading" option)

e.g. "--xpi-out-path"

kumar303:
Since https://github.com/mozilla-jetpack/jpm/pull/495 has landed (which fixes localization), your original patch for each feature will need to be re-submitted as separate patches. I'm closing this one for now but let us know how we can help.

PR #420, opened 2015-12-14
Test on recent versions of node.js

nikolas:
Also, I've removed iojs. iojs is not relevant anymore now that those changes have been merged into node 4.x.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
 
 install:
+  - npm install -g npm
   - npm install
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,9 @@ notifications:
 before_install:
   - "export DISPLAY=:99.0"
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
+  - npm install -g npm
 
 install:
-  - npm install -g npm
   - npm install
 
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 sudo: false
 language: node_js
 node_js:
-  - "iojs"
   - "0.10"
-  - "0.12"
+  - "4"
+  - "5"
 env:
   - FIREFOX_VERSION=nightly
 #  - FIREFOX_VERSION=aurora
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,8 @@
 sudo: false
 language: node_js
 node_js:
-  - "iojs"
-  - "0.10"
-  - "0.12"
```
kumar303: aren't these old versions installed on some systems by default? I thought they were on mac os x but I can't find any evidence of that so maybe not.

kumar303: alright, I did some research and can't find any evidence that these are system defaults so they're probably safe to remove.

.travis.yml
```diff
@@ -1,9 +1,8 @@
...
+  - "4.2"
```
kumar303: could this just be `4` to get any patch/bug fixes?

.travis.yml
```diff
@@ -17,6 +16,9 @@ before_install:
   - "export DISPLAY=:99.0"
   - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
 
+install:
```
kumar303: I think you should add `npm install -g npm` (or equivalent) to upgrade npm before installation. I've seen a lot of bugs with older npms on 4.x around dependencies.

kumar303:
Any idea why it can't find the minimatch dep on 5.x?

.travis.yml
nikolas: node 4.x and 5.x, on travis at least, seem to come already with a compatible version of node.js. ( https://travis-ci.org/mozilla-jetpack/jpm/jobs/96813602#L100 )

Upgrading npm is a good idea on some systems but I don't think it's necessary on travis - i'd prefer to use travis's recommended npm version for that version of node.

.travis.yml
nikolas: Good idea, I will try that. It's unclear from the documentation if that's how it resolves. https://docs.travis-ci.com/user/languages/javascript-with-nodejs

nikolas:
I got the same minimatch dep error on Mac OS X with node 5.x. Upgrading to minimatch 2.0.10 fixed it locally but not on travis..

.travis.yml
nikolas: I think ubuntu 14.04's default node.js is 0.10.x. That version is still probably on many server type systems.. we could leave it in travis.yml if it doesn't cause any problems.

.travis.yml
kumar303: well, I think npm 2.x is hosed on node 4. I always run into things like [this](https://github.com/npm/npm/issues/6551) and upgrading npm fixes it. But I guess if you aren't running into trouble yet then there's no reason to upgrade npm. Just beware of it if you hit any odd npm behavior.

.travis.yml
kumar303: ok, that would be a good reason to continue testing on 0.10

.travis.yml
nikolas: i see.. thanks. Maybe i will add that in there then. I use npm 3.x locally these days so i haven't seen those problems.

kumar303:
> I got the same minimatch dep error on Mac OS X with node 5.x. Upgrading to minimatch 2.0.10 fixed it locally but not on travis..

Do you have the same version of npm locally? You could try an upgrade of npm for Travis just to see what happens. I don't see anything obvious in the output of `npm install` though (aside from minimatch not getting installed for whatever reason).

nikolas:
Alright, tests are now passing when using the latest npm. Thanks for the help.

kumar303:
Thanks for the patch!

PR #423, opened 2015-12-15
Require 'when' module in post.js

Mathieuu:
The `postXpi` in `libs/post.js` cannot be called directly like this: `var post = require('jpm/lib/post');` because `when` is not defined in the module scope.

When calling post, the code breaks here: https://github.com/mozilla-jetpack/jpm/blob/master/lib/post.js#L16.

Ready for code review @FayneAldan @kumar303


kumar303:
Thanks. This code never got test coverage for some reason.

PR #427, opened 2015-12-15
Allow users to specify `sign` polling timeout

JasonGhent:
The extension signing process seems to take much much longer than the hardcoded 60sec magic number. This allows user input to supercede it.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -72,6 +72,7 @@ program
   .option("--api-secret <secret>", "addons.mozilla.org API secret string")
   .option("--api-url-prefix <urlPrefix>", "addons.mozilla.org API URL prefix [" + AMO_API_PREFIX + "]", AMO_API_PREFIX)
   .option("--xpi <filePath>", "optional XPI to be signed. By default, an XPI will be built from your working directory")
+  .option("--timeout <milliseconds>", "time to wait for success/failure response")
   .action(function(options) {
     sign(program, options);
   });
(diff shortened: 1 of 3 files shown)


kumar303:
This sounds like a useful feature, thanks. 

It needs test coverage. You could copy a [test like this](https://github.com/JasonGhent/jpm/blob/user-specified-signing-timeout/test/unit/test.sign.js#L244) which is easy to run : `npm test unit`.

When you manually changed the timeout in the code did your add-on eventually finish?

JasonGhent:
@kumar303 How would a test for this run? Set a timeout >60sec and force the test suite to hold on that test until it finishes at >60sec? Set the opt and check that it is set then bail? Other?

The add-on did finish and everything continued as is seemingly expected, including the downloading of the signed XPI.

kumar303:
Thanks for getting this started. This test suite is a bit odd so I just added in the test before merging. I also incremented the default timeout to two minutes. By the way, there are some known issues causing signing to take a longer than expected but they should be fixed soon.

JasonGhent:
@kumar303 are there any plans on the roadmap to add the `post-url` param to `sign` as a way to upload the *signed xpi* to a specified location?

kumar303:
I hadn't ever heard that feature request. What is the use case? To distribute the XPI? If so, I'd think the endpoint would require authentication so that might be better for developers to script themselves. If you're trying to do this for development then you shouldn't need signing for that. You can use a pre-production Firefox and turn off the signing requirement.

PR #428, opened 2015-12-16
update lodash to 3.10.1

nikolas:


kumar303:
Hmm, it's possible that this test failed because the JWT uses a timestamp for comparison. I can't think of how lodash changes would affect that. We'll see how the restart goes.

nikolas:
weird, well it's passing after a rebase.

kumar303:
ok, it is probably a flaky test then

muhawo:
Lodash`s version should be update.

>npm WARN unmet dependency .\node_modules\grunt-jpm\node_modules\jpm requires lodash@'3.10.1' but will load
npm WARN unmet dependency .\node_modules\lodash,
npm WARN unmet dependency which is version 4.11.1

some other modules will upgrage lodash autoly
>npm WARN deprecated lodash@2.1.0: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^4.0.0.

kumar303:
@muhawo can you recreate a new pull request for the update?

PR #431, opened 2015-12-16
Add test for the "jpm version" command

nikolas:


kumar303:
I didn't write any of the existing functional tests and personally I think they take the wrong approach. They are too heavyweight and too prone to flakiness. Also, I think it is possible to achieve the same effectiveness with a unit test. Could you try it as a unit test? You could do mock injection for stdout and process [just like this test](https://github.com/mozilla-jetpack/jpm/blob/master/test/unit/test.sign.js#L810). It will require some modification of the version command handler but that's ok.

nikolas:
I agree.. this seems like not the best way to write this test. I will re-write it as a unit test.

My motivation for making this test was: we could ensure that, for example, new versions of the commander library work with jpm. Maybe there's a way to test that functionality in another way?

kumar303:
Actually, on second thought (and as you mentioned here), a functional test probably makes sense for this so that it covers the commander. In an ideal world I'd like to see the commander wire-up module to be made more unit testable. I'll merge it as is, thanks.

Issue #258, opened 2015-02-03
Use travis cache to save and reuse Fx on a nightly basis


erikvold:
Since we only use nightly, we can cache on download of it each in travis builds to avoid the 20+ secs it takes to download and unzip this for each build.

nikolas:
Is it still the case that we can only use nightly?

kumar303:
The tests only use nightly https://github.com/mozilla-jetpack/jpm/blob/master/.travis.yml#L27 That could be fixed by adding the variables to the download command but I personally don't think it's worthwhile to introduce more versions (which will make the tests take even longer).

PR #434, opened 2015-12-16
Use FF beta instead of nightly

nikolas:
I'm not sure we need to be manually downloading and installing firefox on travis, since travis already provides a mechanism for using firefox. It's possible we could simplify the travis.yml file like this.

See issue #258

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,32 +4,17 @@ node_js:
   - "0.10"
   - "4"
   - "5"
-env:
-  - FIREFOX_VERSION=nightly
-#  - FIREFOX_VERSION=aurora
-#  - FIREFOX_VERSION=beta
-#  - FIREFOX_VERSION=release
-
+addons:
+  firefox: latest-beta
 notifications:
   irc: "irc.mozilla.org#jpm"
-
 before_install:
-  - "export DISPLAY=:99.0"
-  - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
   - npm install -g npm
-
 install:
   - npm install
-
-before_script:
-  - pwd
-  - cd ..
-  - wget "https://download.mozilla.org/?product=firefox-nightly-latest&os=linux64&lang=en-US" -O firefox.tar.bz2 && tar xvf firefox.tar.bz2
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -4,31 +4,16 @@ node_js:
...
-#  - FIREFOX_VERSION=release
-
+addons:
+  firefox: latest-beta
 notifications:
   irc: "irc.mozilla.org#jpm"
-
 before_install:
-  - "export DISPLAY=:99.0"
-  - "/sbin/start-stop-daemon --start --quiet --pidfile /tmp/custom_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1280x1024x16 -extension RANDR"
```
kumar303: I don't know what this old command was for but it seemed deliberate. Any specific reason for simplifying it?

nikolas: It looks like that command originally came from documentation like this: https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI

There it lists the simplified version - `sh -e /etc/init/xvfb start` - and the complicated one, which seems to be necessary to set a specific screen resolution. I don't have any specific reason for simplifying this command.. I can replace the simplified version with that one.

kumar303:
> I'm not sure we need to be manually downloading and installing firefox on travis

I'm not sure either! I think it was historic for when jpm depended on nightly for features. Nowadays I think it would be safe to run the tests on beta.

.travis.yml
```diff
@@ -4,31 +4,16 @@ node_js:
...
+  - sh -e /etc/init.d/xvfb start
   - npm install -g npm
-
 install:
   - npm install
-
-before_script:
-  - pwd
-  - cd ..
-  - wget "https://download.mozilla.org/?product=firefox-nightly-latest&os=linux64&lang=en-US" -O firefox.tar.bz2 && tar xvf firefox.tar.bz2
```
kumar303: This is an odd request but can you leave this command commented out prefixed with something like 'this is how you get nightly' ? Unfortunately, it is not documented very well and I'm worried we'll forget how to do it if we need to switch back to nightly.

.travis.yml
kumar303: hmm, maybe there was something needed with the screen resolution. I'm not too familiar with the functional tests but I guess putting the old command back would be prudent.

kumar303:
This looks good to go but I think the jshint merge added conflicts

nikolas:
Okay, I fixed that merge conflict

kumar303:
Thanks! This should speed up the builds

Issue #86, opened 2014-07-11
Create a node API


jsantell:
This'll probably have several components to it, but a basic API for using this via node should be possible:

``` javascript
var jpm = require("jpm");
jpm.xpi("/path/to/dir").then(function (xpi) {})
jpm.run("/path/to/dir").then(function () {})
```

Similar to #5 which is having this work in an addon itself, but first got to get the API working and usable before we think about stripping down dependencies for use within an Addon.

PR #437, opened 2015-12-17
Expose sign API

yan12125:
Ref:  #86, #327

With this patch, I can sign my XPI programmatically:
```
var sign = require('/path/to/jpm').sign;
var fs = require('fs');

fs.readFile('./secrets.json', {encoding: 'utf-8'}, function (err, data) {
    if (err) {
        throw err;
    }
    var secrets = JSON.parse(data);
    sign({
        apiKey: secrets.key,
        apiSecret: secrets.secret,
    }).then(function (result) {
        console.log(result);
    }, function (err) {
        throw err;
    });
});
```
Testing:
```
$ node --stack_trace_limit=100 signXpi.js
JPM [warning] Using existing bootstrap.js. This file is usually auto-generated.
Creating XPI
JPM [info] XPI created at /tmp/tmp-unsigned-xpi-72008qSiM5Hk1m50/@newtabnews-0.0.10.xpi (84ms)
Created XPI at /tmp/tmp-unsigned-xpi-72008qSiM5Hk1m50/@newtabnews-0.0.10.xpi
JPM [info] Created XPI for signing: /tmp/tmp-unsigned-xpi-72008qSiM5Hk1m50/@newtabnews-0.0.10.xpi
Validating add-on [............................................................................................]JPM [info] Validation results: https://addons.mozilla.org/en-US/developers/upload/(some hex value)
Downloading signed files: 100% 
JPM [info] Downloaded:
JPM [info]     ./newtab-0.0.10-fx.xpi
{ success: true }
```
Tested with node 5.3.0 on Arch Linux.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -12,7 +12,7 @@ var open = require("open");
 var postXPI = require("../lib/post");
 var watch = require("node-watch");
 var cp = require('child_process');
-var sign = require("../lib/sign");
+var signCmd = require("../lib/sign").signCmd;
 var _ = require("lodash");
 
 var AMO_API_PREFIX = require("../lib/settings").AMO_API_PREFIX;
@@ -75,17 +75,7 @@ program
   .option("--xpi <filePath>", "optional XPI to be signed. By default, an XPI will be built from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
   .action(function(options) {
-    cmd.validateProgram(program);
-    sign(_.assign(options, program)).then(function (result) {
-      console.log(result.success ? 'SUCCESS' : 'FAIL');
-      process.exit(result.success ? 0 : 1);
-    }, function (err) {
-      console.error('FAIL');
-      if (err) {
-        console.error(err.stack);
-      }
-      process.exit(1);
-    });
+    signCmd(program, options);
   });
 
 program
--- a/bin/jpm
+++ b/bin/jpm
@@ -13,7 +13,6 @@ var postXPI = require("../lib/post");
 var watch = require("node-watch");
 var cp = require('child_process');
 var signCmd = require("../lib/sign").signCmd;
-var _ = require("lodash");
 
 var AMO_API_PREFIX = require("../lib/settings").AMO_API_PREFIX;
 // TODO: use .jpmignore file here instead
--- a/bin/jpm
+++ b/bin/jpm
@@ -13,8 +13,9 @@ var postXPI = require("../lib/post");
 var watch = require("node-watch");
 var cp = require('child_process');
 var sign = require("../lib/sign");
+var _ = require("lodash");
 
-var AMO_API_PREFIX = "https://addons.mozilla.org/api/v3";
+var AMO_API_PREFIX = require("../lib/settings").AMO_API_PREFIX;
 // TODO: use .jpmignore file here instead
 var IGNORE_FILES_REGEX = /(?:bootstrap\.js|install\.rdf|\.xpi)$/i;
 
@@ -74,7 +75,17 @@ program
   .option("--xpi <filePath>", "optional XPI to be signed. By default, an XPI will be built from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
   .action(function(options) {
-    sign(program, options);
+    cmd.validateProgram(program);
+    sign(_.assign(options, program)).then(function (result) {
+      console.log(result.success ? 'SUCCESS' : 'FAIL');
+      process.exit(result.success ? 0 : 1);
+    }, function (err) {
+      console.error('FAIL');
+      if (err) {
+        console.error(err.stack);
+      }
+      process.exit(1);
+    });
   });
 
 program
--- a/index.js
+++ b/index.js
@@ -2,5 +2,3 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
-
-module.exports.sign = require('./lib/sign').sign;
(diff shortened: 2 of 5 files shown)


bin/jpm
```diff
@@ -13,8 +13,9 @@ var postXPI = require("../lib/post");
 var watch = require("node-watch");
 var cp = require('child_process');
 var sign = require("../lib/sign");
+var _ = require("lodash");
 
-var AMO_API_PREFIX = "https://addons.mozilla.org/api/v3";
+var AMO_API_PREFIX = require("../lib/settings").AMO_API_PREFIX;
```
kumar303: d'oh, good call. I didn't know about the settings module.

lib/sign.js
```diff
@@ -101,15 +105,6 @@ function sign (program, options, config) {
       return result;
     });
 
-  }).then(function(result) {
-    logger.log(result.success ? 'SUCCESS' : 'FAIL');
-    config.systemProcess.exit(result.success ? 0 : 1);
```
kumar303: because all of these moved, the tests that were checking for process exits failed. I think this patch would be easier f you kept all these as-is but pass in a stub `systemProcess` in your script (I know that's sort of lame but it's easy).

index.js
```diff
@@ -2,3 +2,5 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
+
+module.exports.sign = require('./lib/sign');
```
kumar303: I don't understand why you need this. In your script you should be able to just do this:
````
var sign = require('jpm/lib/sign');
````
Did you try that?

File: bin/jpm [hunk 4]
```diff
@@ -74,7 +75,17 @@ program
   .option("--xpi <filePath>", "optional XPI to be signed. By default, an XPI will be built from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
   .action(function(options) {
-    sign(program, options);
```
kumar303: as mentioned earlier, I think if you put this back then the tests won't need so many changes

kumar303:
btw, you can run all the sign tests in your local code quickly with `npm test unit` (this bypasses the slow functional tests which don't cover the sign command anyway).

lib/sign.js
```diff
@@ -113,5 +124,7 @@ function sign (program, options, config) {
   });
 }
 
-
-module.exports = sign;
+module.exports = {
+    sign: sign,
+    signCmd: signCmd,
```
kumar303: oh nice, splitting them out makes sense

kumar303: I'd suggest instead just for a little less boilerplate:
````
exports.sign = sign;
exports.signCmd = signCmd;
````

**Comment by yan12125** [comment]
Thanks for these comments. The reason I put an export in ```index.js``` is that I believe clients of this package have no knowledge about internal directory structures.
In my third commit, I put ```process.exit``` things and more back to ```sign.js```. I'm just lazy to modify tests :)

File: lib/sign.js [hunk 6]
```diff
@@ -43,10 +50,9 @@ function sign (program, options, config) {
       console.error();
       return reject();
     }
-    cmd.validateProgram(program);
```
kumar303: this actually needs to happen within the promise so that an exception will reject the promise. You could accomplish it by returning a promise in signCmd and chaining them together, like this:
````
function signCmd (program, options, config) {
  return when.promise(function() {
      cmd.validateProgram(program);
    })
    .then(function() {
      return sign(_.assign(options, program), config);
    })
    .then(function (result) {
      logger.log(result.success ? 'SUCCESS' : 'FAIL');
      // ...
    });
}
````

**Comment by kumar303** [comment]
> The reason I put an export in index.js is that I believe clients of this package have no knowledge about internal directory structures

I tried it and it works. I did:
````
mkdir script-test
cd script-test
npm install jpm
node
> sign = require('jpm/lib/sign');
[Function]
````

> I'm just lazy to modify tests :)

We can't accept patches that break the tests for obvious reasons. I think you're on the right path to getting the tests to pass though.

**Comment by yan12125** [comment]
I know ```require('/path/to/jpm/lib/sign')``` works. However I guess scripts other than ```index.js``` are for internal uses only. I have little experience in writing node.js modules, so please correct me if I'm wrong.

Another question: should I squash all commits? The git log looks messy now.

File: index.js [hunk 7]
```diff
@@ -2,3 +2,5 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
+
+module.exports.sign = require('./lib/sign').sign;
```
kumar303: I'd rather not include this unless you add test coverage. I still don't think it's necessary though. It's safe enough to `require('jpm/lib/sign')` in your script

lib/sign.js
```diff
@@ -101,7 +106,16 @@ function sign (program, options, config) {
...
-  }).then(function(result) {
+  });
+}
+
+
+function signCmd (program, options, config) {
+  return when.promise(function(resolve, reject) {
+    resolve(cmd.validateProgram(program));
+  }).then(function () {
+    return sign(_.assign(options, program), config);
```
kumar303: for safety, this should be `_.assign({}, options, program)` so that it doesn't modify the input object.

lib/sign.js
```diff
@@ -24,6 +26,11 @@ function sign (program, options, config) {
     AMOClient: DefaultAMOClient,
```
kumar303: `config.systemProcess` is no longer used in this function so you need to set the default for it in `signCmd` instead. Actually, the `jpm sign` command won't function without this change.

At the top of the `signCmd` just do something similar:
````
config = _.assign({
  systemProcess: process,
}, config);
````
This allows the tests to replace `systemProcess` but the real command to use node's magic `process` variable.

kumar303:
> should I squash all commits?

You can if you want to but it's not that important.

yan12125:
Thanks for reviewing. Updated.

kumar303:
looks great, thanks

Issue #206, opened 2014-10-19
Use mozilla-runner to start firefox


erikvold:
https://github.com/mozilla-b2g/mozilla-runner

erikvold:
I made https://github.com/erikvold/fx-runner instead

anaran:
@erikvold 
> I made https://github.com/erikvold/fx-runner instead
is a 404 now.

I also get

```
~/tmp/mozilla/jpm$ npm install
npm WARN package.json fx-runner@0.0.7 No repository field.
~/tmp/mozilla/jpm$
```

chitcharonko:
I'm having the same problem here...

msangel:
Up

kumar303:
The repository has been deleted and I don't know of any mirrors so we'll need a new patch for this I guess.

kumar303:
oh, my bad, the source had already moved to https://github.com/mozilla-jetpack/node-fx-runner

kumar303:
looks like the repository field was already updated too: https://github.com/mozilla-jetpack/node-fx-runner/blob/master/package.json#L19

kumar303:
I'll try to upgrade jpm's dep version here https://github.com/mozilla-jetpack/jpm/pull/438

PR #438, opened 2015-12-17
Upgrade fx-runner

kumar303:


kumar303:
I manually tested `jpm run` and `jpm test` which both seemed to work.

kumar303:
see https://github.com/mozilla-jetpack/jpm/issues/206

Issue #440, opened 2015-12-18
jpm xpi generates invalid .update.rdf


benmccann:
This line is broken:
https://github.com/mozilla-jetpack/jpm/blob/master/lib/rdf.js#L158

The tool is exporting:

    <Description about="urn:mozilla:myext@myext.com">

When it should be exporting:

    <Description about="urn:mozilla:extension:myext@myext.com">

No users of our extension got the update to the latest version of our extension because of this bug :-(  Which meant they all have unsigned versions installed, which turned out to make this an awesome day with FF43 rolling out.

https://developer.mozilla.org/en-US/docs/Extension_Versioning,_Update_and_Compatibility

>It is important to get the initial RDF:Description's about attribute correct. It varies depending on what type of add-on you are providing information for:
>
>* For an extension or plugin bundle it must be urn:mozilla:extension:<id>
* For a theme it must be urn:mozilla:theme:<id>
* For any other type of add-on it must be urn:mozilla:item:<id>

benmccann:
@guyzmuch @kumar303 either of you guys know if we can just hardcode an extra "extension:" to fix this? jpm isn't used to build themes or other types of add-ons is it?

kumar303:
Huh. The about formatting (without `:extension`) has been the same since it was added [back in May](https://github.com/mozilla-jetpack/jpm/commit/9b31b885). Did something change in Firefox?

Either way, if there's a fix, I'd be happy to get it in. I don't think jpm is used for themes but I also don't know much about how updates work.

@rpl do you have some familiarity here to help make a decision on how to fix the `about` property?

benmccann:
He wrote the code back in May, but no one got around to merging it until October https://github.com/mozilla-jetpack/jpm/commit/004f02915b748345a95e2068b64c02178bda0b14. This is the first release we've cut with `jpm` instead of `cfx`. I'm pretty sure it was just broken when initially written and merged

kumar303:
oh, I see.

benmccann:
Sent a PR for this: https://github.com/mozilla-jetpack/jpm/pull/441

kumar303:
Thanks again

benmccann:
Thanks for being so responsive!  If you know anyone on the addon-sdk I have some bug fixes from August I'm still trying to get merged :-)

kumar303:
[...]

PR #441, opened 2015-12-19
Fix RDF:Description's about attribute

benmccann:


kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/440

kumar303:
Thanks for the patch. A big change landed to clean up some code lint, can you rebase on master? Also, is there an easy way to add a test for this? I haven't looked at the RDF tests so maybe it's not so easy.

benmccann:
Rebased. I don't think it'd be too hard to add a test, but I'm not very familiar with npm development and I don't think I'll get to it for awhile. Maybe we could merge this, but leave the bug open until the test is added?

PR #442, opened 2015-12-19
Validate JS code style with jscs

nikolas:
I've set it up to check everything in lib/ automatically.
There are still many failures in the test/ directory.
Also, I've turned off jsDoc checking for now because it
was complaining about some functions having documented
parameters that it looks like jpm is parsing via some
"arguments" variable, not through the function signature.

### DIFF
--- a/.jscsrc
+++ b/.jscsrc
@@ -1,0 +1,7 @@
+{
+  "preset": "google",
+  "validateIndentation": 2,
+  "validateQuoteMarks": "\"",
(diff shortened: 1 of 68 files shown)


kumar303:
Wow, thanks for slogging through all this! I scanned through it and my eyeballs hurt but nothing jumped out at me. Doing a first pass without linting the tests is fine. We have jshint which is the main thing to catch unused vars and that sort of thing.

PR #447, opened 2015-12-21
Restore default behavior of getManifest()

kumar303:
The default behavior regressed in https://github.com/mozilla-jetpack/jpm/pull/399, see https://github.com/mozilla-jetpack/jpm/commit/bb20e034a921f76f1ae6af7cfefc6697f3e34d36#commitcomment-15107293

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -88,13 +88,15 @@ exports.console = Object.freeze(jpmConsole);
  * from this directory after a temporary XPI extraction.
  *
  * @param {Object} options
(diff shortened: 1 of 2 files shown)


FayneAldan:
Btw, this issue caused my build script to stop working, in turn delaying the approval of my minimized addon for too long... Try not to let this happen again... :|

kumar303:
We are unable to guarantee the stability of each function when used directly via scripts. For more stability I'd suggest automating the command line interface and / or helping add missing feature for what you need in the command line interface.

PR #448, opened 2015-12-22
inject bootstrap.js and install.rdf directly in the generated xpi file

rpl:
This PR introduces the minimal changes needed to inject the fallback bootstrap.js and install.rdf files
in the generated xpi instead of generating the temporary files in the addon dir and then removing them.

### DIFF
--- a/lib/xpi/utils.js
+++ b/lib/xpi/utils.js
@@ -6,6 +6,8 @@
 var join = require("path").join;
 var when = require("when");
 var fs = require("fs-promise");
+var Zip = require("jszip");
+
 var RDF = require("../rdf");
 var zip = require("../zip");
 var console = require("../utils").console;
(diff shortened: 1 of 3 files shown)


lib/xpi/utils.js
```diff
@@ -6,6 +6,8 @@
 var join = require("path").join;
 var when = require("when");
 var fs = require("fs-promise");
+var Zip = require("jszip");
```
kumar303: ooh, can we use `jszip` for other things? I tried to fix another issue [like this](https://github.com/mozilla-jetpack/jpm/pull/410) but it broke tests in mysterious ways. Maybe I'll try to resurrect it with `jszip`.

lib/xpi/utils.js
```diff
@@ -75,34 +77,24 @@ function createFallbacks(options) {
     console.log("Creating fallbacks if they are necessary..");
   }
 
-  return getManifest({
-    addonDir: options.addonDir
-  }).then(function(manifest) {
+  return getManifest({addonDir: options.addonDir}).then(function(manifest) {
```
kumar303: are you rebased on the latest master? I think this indentation change was needed for the newly added lint checks (probably for line length).

lib/xpi/utils.js
```diff
@@ -112,6 +104,35 @@ function createZip(options) {
...
+    if (!fallbacks || (!fallbacks.installRDF && !fallbacks.bootstrapJS)) {
+      // return the zip file if there isn't any needed fallback file
+      return result;
+    }
+
+    return fs.readFile(options.xpiPath).then(function(data) {
+      // add the needed fallbacks files into the zip file
+      var zip = new Zip(data);
+
+      console.log("Inject bootstrap.js and install.rdf in " +
```
kumar303: was this just for debugging? or is it something that should be displayed while jpm is running?

kumar303: if so, maybe wrap it in `if options.verbose ...` ?

kumar303:
Looks straight forward. Do you think the current tests cover this well enough? I guess it's all the same logic as before, right?

I'm not sure why Travis didn't run the tests for this pull request. Hmm.

lib/xpi/utils.js
rpl: @kumar303 I choose ```jszip``` mostly because it is what ```zip-dir``` uses internally.
I'm not sure about the issues on #410 but it can definitely used to directly manipulate zip files.

rpl:
@kumar303 it seems to be already covered by unit tests:
- https://github.com/mozilla-jetpack/jpm/blob/master/test/unit/test.xpi.js#L143
- https://github.com/mozilla-jetpack/jpm/blob/master/test/unit/test.xpi.js#L171
- https://github.com/mozilla-jetpack/jpm/blob/master/test/unit/test.xpi.js#L198

package.json
```diff
@@ -57,7 +57,8 @@
     "tmp": "0.0.28",
     "unzip": "0.1.11",
     "when": "3.7.2",
-    "zip-dir": "1.0.0"
+    "zip-dir": "1.0.0",
+    "jszip": "^2.4.0"
```
kumar303: oh I just noticed the caret. I think we want to pin all dependencies to their exact version, like `"jszip": "2.4.0"` because otherwise someone cloning the repo for the first time may encounter a bug from using a newer version than what the existing developers were using. I know theoretically the caret should only pick up bug fixes but that can't be fully trusted :)

rpl: just pushed an updated commit with the fixed package.json

kumar303:
:+1: :zap:

Issue #450, opened 2015-12-22
jpm sign: Version does not match install.rdf. ( status: 400 )


pnguyen-pi:
Hi everybody,

I tried to use "jpm sign" command to retrieve a new XPI for my add-on signed by Mozilla. But I got a strange issue regard to version number does not match install.rdf file

> JPM [info] Signing XPI: 
> JPM [error] Server response: Version does not match install.rdf. ( status: 400 )
> JPM [info] FAIL

I verified the install.rdf file and its version is 2.14.7. Also tried to installed this XPI to my FF browser and it still displayed version of 2.14.7 in the FF browser. Not sure how I got this issue and how to bypass it because I also tried to create a new submission of this XPI manually on Develop Hubs and no issue/warning is returned like above.

Please help me on this issue.

kumar303:
Hi. I think jpm may be detecting the version wrong. Is it possible to upload the XPI file you were trying to sign? Were you running it like `jpm sign --xpi=path/to/the.xpi` ?

kumar303:
Aha. I found a file with an alternate RDF that reproduces this bug.

(grrr, github won't let me attach an XPI).

kumar303:
The code is looking for a manifest in the XPI, not finding it, and posting to `/api/v3/addons/null/versions/undefined/`, d'oh. Yay javascript.

pnguyen-pi:
@kumar303  I used the command as below to get signning from Mozilla

>jpm sign --api-key "user:11853170:122" --api-secret "secret" --xpi "path/to/the.xpi"

I am looking forward your fix (^.^)

Thank you so much.

zeird:
> The code is looking for a manifest in the XPI, not finding it, and posting to /api/v3/addons/null/versions/undefined/, d'oh. Yay javascript.

So would the next versions of `jpm sign` ignore the absence of version number in package.json?

notnarb:
As suggested in https://github.com/mozilla-jetpack/jpm/pull/456 this (at least for me) is caused by a lack of a package.json within the xpi (mine was created with crx).  

A quick and dirty fix that worked for me was to just ```zip <filename>.xpi package.json``` as part of my build process since I can't seem to find a way to inform crx that I wish to have the package.json included.

kumar303:
I think this has been fixed as part of #465. Could you test it out following these [install instructions](https://github.com/mozilla-jetpack/jpm/#install) to see if it works before we publish a release?

[...]

PR #456, opened 2015-12-30
Show a better error for non SDK-based XPIs

kumar303:
This should get us better bug reports, instead of https://github.com/mozilla-jetpack/jpm/issues/450

This also makes progress toward parsing `install.rdf` files.

You see a more sensible error:
````
$ jpm --verbose sign --api-key user:1:458 --api-secret ... --api-url-prefix http://olympia.dev/api/v3 --xpi ~/tmp/addons/alt-rdf.xpi
JPM [info] verbose set
JPM [error] FAIL
Error: XPIs without a package.json are not supported yet
    at /Users/kumar/dev/jpm/lib/sign.js:45:17
    at tryCatchReject (/Users/kumar/dev/jpm/node_modules/when/lib/makePromise.js:845:30)
    at runContinuation1 (/Users/kumar/dev/jpm/node_modules/when/lib/makePromise.js:804:4)
    at Fulfilled.when (/Users/kumar/dev/jpm/node_modules/when/lib/makePromise.js:592:4)
    at Pending.run (/Users/kumar/dev/jpm/node_modules/when/lib/makePromise.js:483:13)
    at Scheduler._drain (/Users/kumar/dev/jpm/node_modules/when/lib/Scheduler.js:62:19)
    at Scheduler.drain (/Users/kumar/dev/jpm/node_modules/when/lib/Scheduler.js:27:9)
    at doNTCallback0 (node.js:419:9)
    at process._tickCallback (node.js:348:13)
````

### CODE REVIEW

File: lib/utils.js [hunk 1]
```diff
@@ -115,7 +116,41 @@ function getManifest(options) {
...
+ *
+ * The resolution object has the following attributes:
+ * - `path`: temporary directory path containing the XPI contents.
+ * - `remove()`: method that removes the temporary directory. You must
+ *   call this explicitly when you are finished.
+ *
+ * @param {Object} tmpXPI
+ * @return {Promise}
+ */
+function inspectXPI(xpiPath) {
```
mstriemer: "inspect" doesn't sound right. This actually extracts it, right? I would imagine `inspectXPI` would return a size and file list or something. Perhaps `extractXPI` or something else is a better name?

kumar303: heh, I had it named that originally. Names are hard. My thinking was that since it's temporary you do extract it but the main purpose is for inspecting. I can change it back though, doesn't really make sense now that I try to explain it.

lib/sign.js
```diff
@@ -17,11 +18,51 @@ var xpi = require("./xpi");
...
+    xpiPath: null,
+  }, options);
+
+  if (!options.xpiPath) {
+    return utils.getManifest({addonDir: options.addonDir})
+      .then(function(manifest) {
+        return getXpiInfoFromManifest(manifest);
+      });
+  } else {
+    return utils.inspectXPI(options.xpiPath)
```
mstriemer: Looks like this could be:

```js
return utils.getXPIManifest(options.xpiPath)
    .then(function(manifest) {
      if (manifest) {
        return getXpiInfoFromManifest(manifest);
      } else {
        throw new Error("XPIs without a package.json are not supported yet");
      }
    });

File: lib/sign.js [hunk 3]
```diff
@@ -17,11 +18,51 @@ var xpi = require("./xpi");
...
+    xpiPath: null,
+  }, options);
+
+  if (!options.xpiPath) {
+    return utils.getManifest({addonDir: options.addonDir})
+      .then(function(manifest) {
+        return getXpiInfoFromManifest(manifest);
+      });
+  } else {
+    return utils.extractXPI(options.xpiPath)
```
mstriemer: Looks like this could be:

```js
return utils.getXPIManifest(options.xpiPath)
    .then(function(manifest) {
      if (manifest) {
        return getXpiInfoFromManifest(manifest);
      } else {
        throw new Error("XPIs without a package.json are not supported yet");
      }
    });
```

(I deleted the old comment because it got "resolved" immediately with the rename)

kumar303: it could but when I get around to implementing the RDF parsing, the else branch would have parsed the XPI twice which seems wasteful -- i.e. a non-manifest XPI will require a call to `extractXPI()`

mstriemer: It doesn't feel right to duplicate loading the package.json file but I guess it isn't much code. You could extract that into a method and make this just set `var manifest = utils.loadManifest(tmpXPI)`. It's small but it might start getting duplicated in lots of places.

mstriemer: Actually `getXPIManifest` doesn't have anything specific to a `package.json` file in there. You could potentially load an `install.rdf` as a fallback there if all you need is data they both have.

mstriemer:
I think something's wrong, the tests passed.

I'd lean towards reusing `getXPIManifest` and changing it later if you can't put the logic in there but I'm fine with it this way too.

r+

lib/sign.js
kumar303: Ah, thanks, I took a closer look and I think I can merge the two. It seems that `sign` is the only one that passes `xpiPath` to `getManifest()` so I can change its behavior. Before, I was scared to change how it handled missing manifest files :fearful:

kumar303:
I fixed it up to re-use the other `package.json` parsing code.

kumar303:
The reason I didn't merge it all into `getManifest()` is that `install.rdf` isn't really a "manifest" like `package.json`, they are different.

Issue #465, opened 2016-01-17
Support non-SDK addons with jpm sign?


willsALMANJ:
Right now, `jpm sign` parses the `package.json` file in order to determine an addon's metadata. In https://github.com/mozilla-jetpack/jpm/pull/456, progress towards parsing `install.rdf` is mentioned (and the error message "These XPIs are not supported yet" was added for non-SDK addons). All that is needed to sign an addon is to parse `install.rdf` for the addon id and version (plus the AMO credentials and `.xpi` file).

Is supporting non-SDK addons something that is planned to be added in the near future, or just something that would be nice to add but with no immediate plans? It is slightly redundant, but I wouldn't mind `jpm sign` working on non-SDK addons by requiring that the addon id and version be passed as command line parameters rather than parsing the `install.rdf` file if that were easier to implement.

For now, I use my own script (https://github.com/willsALMANJ/pentadactyl-signed/blob/master/amo_xpi_sign.py) for signing old XUL addons, but it would be nice to `jpm` instead since it is better tested and maintained.

kumar303:
Non SDK addons will definitely be supported soon, the RDF parsing just
hasn't been implemented yet.

willsALMANJ:
Thanks for the response! I will watch this to know when signing non-SDK addons has been added.

kumar303:
It should be fixed on master. Could you test it out following these [install instructions](https://github.com/mozilla-jetpack/jpm/#install) to see if it works before we publish a release?

willsALMANJ:
Works for me, thanks!

kumar303:
![highfive](https://cloud.githubusercontent.com/assets/55398/12528128/19e34f44-c153-11e5-8fa6-c2170768dd99.gif)

meabhi007:
I am using the installation instructions provided to install JPM using github.
Still getting the error: JPM [error] Server response: Duplicate add-on ID found. ( status: 400 )
Do i need to modify <em:id> tag in install.rdf file to upload the revised version of extension?
Tried changing version in package.json, but it didn't help.

kumar303:
Yes, it looks for `package.json` first then falls back to `install.rdf`. Both need to be up to date with the correct information.

PR #466, opened 2016-01-21
jpm sign: support non-SDK add-ons via install.rdf

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/465


kumar303:
@mstriemer r?

lib/utils.js
```diff
@@ -121,11 +123,13 @@ function getManifest(options) {
 exports.getManifest = getManifest;
 
 function getXpiInfoFromManifest(manifest) {
-  return {
-    manifest: manifest,
-    id: getID(manifest),
-    version: manifest.version,
-  };
+  return when.promise(function(resolve) {
```
mstriemer: I think this could just be:

```js
return when({
  manifest: manifest,
  ...
});

File: lib/utils.js [hunk 2]
```diff
@@ -141,16 +145,34 @@ exports.getXpiInfoFromManifest = getXpiInfoFromManifest;
...
-      // TODO: when there's no manifest, parse install.rdf to get info.
+      if (fileExists(packageFile)) {
+        getInfo = getXpiInfoFromManifest(require(packageFile));
+      } else if (fileExists(installRdfFile)) {
+        getInfo = nodefn.call(fs.readFile, installRdfFile)
+          .then(function(data) {
+            return getXpiInfoFromInstallRdf(data);
+          });
+      } else {
+        getInfo = when.promise(function() {
```
mstriemer: I think you could use `when.reject(new Error("Cannot get info: no manifest found in this XPI"));` here.

lib/utils.js
```diff
@@ -141,16 +145,34 @@ exports.getXpiInfoFromManifest = getXpiInfoFromManifest;
...
-        manifest = require(path.join(tmpXPI.path, "package.json"));
-      } catch (e) {}
-
-      tmpXPI.remove();
+      var getInfo;
+      var packageFile = path.join(tmpXPI.path, "package.json");
+      var installRdfFile = path.join(tmpXPI.path, "install.rdf");
 
-      // TODO: when there's no manifest, parse install.rdf to get info.
+      if (fileExists(packageFile)) {
```
mstriemer: I feel like there could be cases where there is a package.json but it isn't a jpm.

Is there something in the file you could check to verify it is a jpm and have this fail if the file doesn't exist or it isn't a jpm?

I'm thinking something like:

```js
getInfo = getXpiInfoFromManifest(require(packageFile))
  .catch(ManifestError, function(err) {
      return getXpiInfoFromInstallRdf(data);
  }).catch(ManifestError, function(err) {
      throw new Error("Cannot get info: no manifest found in this XPI");
  });

Comment (mstriemer) [inline lib/utils.js]: If it isn't a jpm it might have a package.json just to install jpm with `npm install`, for example. But there could be dependencies that get bundled as well.

File: lib/utils.js [hunk 4]
```diff
@@ -203,3 +225,65 @@ function extractXPI(xpiPath) {
...
+      var foundManifest = false;
+      descriptions.forEach(function(obj) {
+        if (obj.$["RDF:about"] !== "urn:mozilla:install-manifest") {
+          return;
+        }
+        foundManifest = true;
+        // Check properties or attributes:
+        info.id = obj["em:id"] ? obj["em:id"][0] : null;
+        if (!info.id) {
+          info.id = obj.$["em:id"];
```
mstriemer: What's the difference between `obj["em:id"]` and `obj.$["em:id"]`?

lib/utils.js
kumar303: ooh, I didn't know about that

lib/utils.js
kumar303: The XML 2 object thing puts attributes in a magical `$`

`¯\_(ツ)_/¯`

kumar303: I guess I could explain that in the comment above.

mstriemer:
Looks good, r+wc.

lib/utils.js
kumar303: hmm, I like the idea but it throws wrenches in a couple other places. I don't know if I have the stomach for it on this patch.

Issue #470, opened 2016-01-25
jpm sign might download unsigned files for custom XPIs


kumar303:
When trying to sign an XPI without a compatible application (`targetApplication` in install.rdf), jpm downloads unsigned files and calls it a "success" which is wrong and confusing.

Running `jpm sign` on [invalid-app-ver.zip](https://github.com/mozilla-jetpack/jpm/files/103531/invalid-app-ver.zip) (change name to .xpi):

````
Downloading signed files: 100%
JPM [info] Downloaded:
JPM [info]     ./bug_541420-1.0.xpi
JPM [info] SUCCESS
$ unzip -l bug_541420-1.0.xpi
Archive:  bug_541420-1.0.xpi
  Length     Date   Time    Name
 --------    ----   ----    ----
      619  01-25-16 10:42   install.rdf
 --------                   -------
      619                   1 file
````
As you can see, it downloaded an unsigned XPI. The [validation results](https://addons-dev.allizom.org/en-US/developers/upload/66c04be8c5bf4e18bc8e74d3df60c031) for this one I just tried say that it's valid but I think something in AMO is skipping signing because of the missing target application.

kumar303:
Huh. Here is another XPI (which is missing an `install.rdf`) that causes jpm to download unsigned files. 
[no-install-rdf.zip](https://github.com/mozilla-jetpack/jpm/files/103548/no-install-rdf.zip) (change name to .xpi).

````
$ unzip -l my_simple_addon-1.0.0.xpi
Archive:  my_simple_addon-1.0.0.xpi
  Length     Date   Time    Name
 --------    ----   ----    ----
       29  02-11-14 15:00   index.js
      204  02-11-14 15:00   package.json
 --------                   -------
      233                   2 files
````

PR #472, opened 2016-01-26
Don't download unsigned files

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/470

If no files were signed at all then you will get an informative error with a traceback.

If there some files were signed but not all of them (I think this case is highly unlikely), jpm will just show a warning message and will download just the signed files.

This addresses a legitimate case where add-ons that do not target Firefox do not get signed (because why would you need them signed).

### DIFF
--- a/test/unit/test.sign.js
+++ b/test/unit/test.sign.js
@@ -79,6 +79,7 @@ describe("amoClient.Client", function() {
         valid: true,
         reviewed: true,
         files: [{
+          signed: true,
           download_url: "http://amo/some-signed-file-1.2.3.xpi",
         }],
         validation_url: "http://amo/validation-results/",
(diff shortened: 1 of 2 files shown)


test/unit/test.sign.js
```diff
@@ -360,6 +362,76 @@ describe("amoClient.Client", function() {
...
+        request: fakeRequest.getCallable(),
+        createWriteStream: createWriteStream.getCallable(),
+        stdout: {
+          write: function() {},
+        }
+      }).then(function(result) {
+        var filePath = path.join(process.cwd(), "some-signed-file-1.2.3.xpi");
+        expect(result.success).to.be.equal(true);
+        expect(result.downloadedFiles).to.be.deep.equal([filePath]);
+        expect(fakeRequest.call[0].url).to.be.equal(files[0].download_url);
```
mstriemer: Would this test make sense `expect(fakeRequest.call.length).to.be.equal(files.length -1)`? I was thinking you'd check that `http://nope.org/should-not-be-downloaded.xpi` wasn't downloaded based on it's URL.

mstriemer:
r+

Test change might be nice but what's there looks good too.

test/unit/test.sign.js
kumar303: oh yeah, good idea

Issue #474, opened 2016-01-27
jpm@1.0.5 does not compress files when packing .xpi (with update.rdf)


jc3213:
Environment: 
Win10 64bit 2015LTSB, nodejs 5.5.0, jpm 1.0.5

Information: 
The result is that the compress ratio of the packed xpi by jpm 1.0.5 is 100%, which means none of the files have been compressed

I've installed jpm 1.0.4 using npm install jpm@1.0.4 -g and then packed my-addon.xpi via jpm xpi. The compress ratio is about 42% (recognized by winrar)

I have install/uninstall nodejs, run npm install jpm -g and npm uninstall jpm -g for many times...but nothing changes, I wonder if there is something wrong with my build environment or this is the bug of jpm 1.0.5?

kumar303:
That seems bizarre. Is the XPI a valid XPI that you can install? Could you attach a zip of the source (if possible) so that I could unzip it and try running `jpm xpi` to reproduce the issue?

jc3213:
Yes, the XPI is valid to install and is able to be signed by jpm sign. and I've asked other people to try that, they also get uncompressed XPI file.

jc3213:
Github keeps warning that "Unfortunately, we don’t support that file type. Try again with a PNG, GIF, JPG, DOCX, PPTX, XLSX, TXT, PDF, or ZIP. "

Would you mind if you can download from https://github.com/jc3213/SimpleProxy/releases/download/2.0.1/myaddon.ZIP here?

kumar303:
Thanks, I downloaded it and will take a look tomorrow.

> the maxVersion is only 43.0, but it is now 44.0 in release

Version incompatibility is just a relic of the past. It's not enforced anywhere in Firefox anymore to my knowledge.

jc3213:
Yes, I know,but it's annoying  when I visit Manage My Submission on AMO

kumar303:
annoying in what way? Do you see a warning? That might be something we could fix easily.

kumar303:
ok, I see what you mean. I can also see 0% compression with `unzip` on an XPI created with jpm from the master repo branch. If I switch to the `1.0.4` tag I see 62% (as expected). Comparison: 

````
$ git checkout master
$ npm install
...
$ pushdir ~/Downloads/Simple\ Proxy
$ jpm xpi
...
$ unzip -lv simpleproxy\@jc3213.github-2.0.2.xpi
[...]

PR #475, opened 2016-01-27
Generating compressed XPI by default and providing compression-level option to users

inbasic:
JPM similar to CFX should generate compressed XPIs by default.

Examples:
```bash
jmp xpi --compression-level 0 # no compression 
jmp xpi --compression-level 9 # best compression
jpm xpi # compression level is 5 by default
```

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -39,6 +39,7 @@ program
   .option("--binary-args <CMDARGS>", "Pass additional arguments into Firefox.")
   .option("--prefs <path>", "Custom set user preferences (path to a json file)")
   .option("--post-url <url>", "A url to post a xpi of your extension")
+  .option("--compression-level <number>", "The compression level of the generated xpi (any level between 0 (no compression) and 9 (best but slowest compression). Default level is 5)")
   .option("--stop-on-error", "Stop running tests after the first failure")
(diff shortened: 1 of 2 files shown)


andymckay:
This seems of limited value of since the xpi is going to be passed to AMO for signing so it can be distributed. That server is going to sign it and rezip it: https://bugzilla.mozilla.org/show_bug.cgi?id=1169463

jc3213:
@andymckay Please have a look into #474 , jpm 1.0.5 doesn't compress xpi when making xpi or signing it. and even the signed xpi by jpm sign is uncompressed. I think the server is not about to repacking all add-ons

kumar303:
> jpm 1.0.5 doesn't compress xpi when making xpi or signing it

that's a separate issue which I think is a regression from special-casing some `update.rdf` code.

kumar303:
What Andy said is true so it may not be very useful (since all XPIs need to be signed now). If you really need the patch then it would at least need some tests. Run `npm test unit` to develop on them. You probably need to add one to https://github.com/inbasic/jpm/blob/master/test/unit/test.xpi.js and I'd suggest using a mock object for `zip` and just make assertions about the constructor args.

inbasic:
I am going to close this pull request as it doesn't seem to have enough value.

PR #478, opened 2016-01-27
Fix #473: Allow arbitrary engines

### LINKED ISSUE #473: Allow arbitrary engines
I suggest allowing arbitrary engines that are formatted as UUID. Those would still be included into the install.rdf. This would allow developers to generated extensions that also work in other not officially supported applications without having to always modify the generated xpi, or maybe even use jpm run with those applications.

freaktechnik:
I've tried to look for places to test thise and only found the direct test of the RDF generation module. I'd expect that to be enough, since the error I've seen when trying to use GUIDs for engines in the past is generated in said module (and will still be generated if an engine is neither a GUID nor a recognized engine name).

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -222,6 +222,9 @@ function createUpdateRDF(manifest) {
 exports.createUpdateRDF = createUpdateRDF;
 
 function getGUID(type) {
+  if (type.match(/^\{[a-f0-9]{8}-(?:[a-f0-9]{4}-){3}[a-f0-9]{12}\}$/)) {
+    return type;
+  }
   return GUIDS[(type = type.toUpperCase())];
 }
 
--- a/test/unit/test.rdf.js
+++ b/test/unit/test.rdf.js
@@ -333,6 +333,16 @@ describe("lib/rdf", function() {
       expect(firefox.childNodes[5].tagName).to.be.equal("em:maxVersion");
       expect(firefox.childNodes[5].childNodes[0].data).to.be.equal(MAX_VERSION);
     });
+
+    it("passes through raw GUIDs as targetApplication", function() {
+      var xml = setupRDF({ engines: {
+        "{e95a4b1e-901d-d035-4077-e95157a7a118}": "*"
+      }});
+      var apps = xml.getElementsByTagName("em:targetApplication");
+      var application = apps[0].childNodes[1]; // Description
+      expect(application.childNodes[1].tagName).to.be.equal("em:id");
+      expect(application.childNodes[1].childNodes[0].data).to.be.equal("{e95a4b1e-901d-d035-4077-e95157a7a118}");
+    });
   });
 
   describe("createUpdateRDF", function() {


lib/rdf.js
```diff
@@ -222,6 +222,8 @@ function createUpdateRDF(manifest) {
 exports.createUpdateRDF = createUpdateRDF;
 
 function getGUID(type) {
+  if (type.match(/^\{[a-f0-9]{8}-(?:[a-f0-9]{4}-){3}[a-f0-9]{12}\}$/))
```
freaktechnik: Looking at the results from travis I should add braces to this if statement. Should I squash the fix?

kumar303: yes, please squash the commits if possible.

freaktechnik:
I don't think the timeout in the Node 5 tests is related?

kumar303:
No, I think the timeout is a known flaky test. I restarted the build :fried_shrimp:

test/unit/test.rdf.js
```diff
@@ -333,6 +333,21 @@ describe("lib/rdf", function() {
       expect(firefox.childNodes[5].tagName).to.be.equal("em:maxVersion");
       expect(firefox.childNodes[5].childNodes[0].data).to.be.equal(MAX_VERSION);
     });
+
+    it("passes through raw GUIDs as targetApplication", function() {
+      var xml = setupRDF({ engines: {
+        "{ec8030f7-c20a-464f-9b0e-13a3a9e97384}": "*"
```
kumar303: I don't think this is such a good test because it's identical to the Firefox GUID. How about just something like `{xxxxxxxx-c20a-464f-9b0e-yyyyyyyyyyyy}` ?

test/unit/test.rdf.js
```diff
@@ -333,6 +333,21 @@ describe("lib/rdf", function() {
...
+      }});
+      var apps = xml.getElementsByTagName("em:targetApplication");
+      var firefox = apps[0].childNodes[1]; // Description
+      expect(firefox.childNodes[1].tagName).to.be.equal("em:id");
+      expect(firefox.childNodes[1].childNodes[0].data).to.be.equal(GUIDS.FIREFOX);
```
kumar303: then you could update this assertion

test/unit/test.rdf.js
```diff
@@ -333,6 +333,21 @@ describe("lib/rdf", function() {
...
+
+      expect(firefox.childNodes[3].tagName).to.be.equal("em:minVersion");
+      expect(firefox.childNodes[3].childNodes[0].data).to.be.equal(MIN_VERSION);
+      expect(firefox.childNodes[5].tagName).to.be.equal("em:maxVersion");
+      expect(firefox.childNodes[5].childNodes[0].data).to.be.equal(MAX_VERSION);
```
kumar303: for this test, I don't think it's important to make any assertions about minVersion or maxVersion. Those are covered in other tests.

kumar303:
Let me know when it's updated and I'll take another look.

freaktechnik:
Okay, I've updated it with an example GUID that does not exist. Possibly. I also removed the version assertions.

kumar303:
Thanks!

PR #484, opened 2016-02-01
Turn addon-dir into an absolute path (fixes #483)

### LINKED ISSUE #483: addon-dir command line option doesn't work with relative paths
Example:

screen shot 2016-02-01 at 11 38 07

I'll submit a patch.

johannhof:
The addonDir option seems to only work fine with absolute paths. This is
fine as long as jpm only uses those internally and for tests, but it
breaks the `jpm xpi --addon-dir ./my/addon` command. The easiest fix is
to just run a transform function on that CLI option that makes it absolute.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -13,6 +13,7 @@ var postXPI = require("../lib/post");
 var watch = require("node-watch");
 var cp = require("child_process");
 var signCmd = require("../lib/sign").signCmd;
+var path = require("path");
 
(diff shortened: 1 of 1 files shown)


johannhof:
I'm not sure how to test this as we don't seem to have tests affecting the command line options. Maybe no tests this time? :pig_nose:

kumar303:
> we don't seem to have tests affecting the command line options

Yes, this is sadly true. I wouldn't suggest refactoring this module to be testable just for this patch :/

kumar303:
thanks!

PR #486, opened 2016-02-02
Run JSHint and JSCS pre-push

johannhof:
This is a really minor thing, but I usually tend to forget to run either jscs or jshint before making a pull request. This helps.


kumar303:
Good idea! Thanks

PR #488, opened 2016-02-02
Actually compress the XPI (for updatable XPIs)

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/474

### DIFF
--- a/lib/xpi/utils.js
+++ b/lib/xpi/utils.js
@@ -129,7 +129,9 @@ function createZip(options) {
         zip.file("bootstrap.js", fallbacks.bootstrapJS);
       }
 
-      var buffer = zip.generate({type: "nodebuffer"});
+      var buffer = zip.generate({type: "nodebuffer",
+                                 compression: "DEFLATE",
+                                 compressionOptions: {level: 6}});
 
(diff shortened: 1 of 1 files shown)


rpl:
@kumar303 looks good, sorry for the regression.

node-zip-dir doesn't currently specify a compression level (but we currently rewrite the xpi even when the fallbacks are not injected, then currently the compression level will be set in both the cases).

Maybe, in a follow up, we can even skip the zip rewriting when no fallback needs to be injected (and adding a test case which checks that the compression level is not 0% can help to prevent future regressions on this)

kumar303:
> node-zip-dir doesn't currently specify a compression level 

You mean internally? I can see that `node-zip-dir` does [specify a compression level here](https://github.com/jsantell/node-zip-dir/blob/master/index.js#L44).

Good ideas for the follow-ups. Thanks for the review.

rpl:
> You mean internally? I can see that node-zip-dir does specify a compression level here.

@kumar303 yep, I meant that it doesn't use ```compressionOptions: {level: 6}``` to select the preferred compression level.

Issue #461, opened 2016-01-11
jpm xpi creates invalid install.rdf


benbasson:
With package.json like so:

    "engines": {
      "firefox": ">=44.0"
    },

jpm xpi creates an install.rdf with the following:

    <em:minVersion>44.0</em:minVersion>
    <em:maxVersion>43.0</em:maxVersion>

Obviously AMO rejects this.

lalakiwe:
I believe the default value of em:maxVersion is not good enough. However, you could specific the maxVersion for your addon alternatively as below:
```
  "engines": {
    "firefox": ">=44.0 || <=49.*"
  },
```
The output in install.rdf will be:
```
<em:minVersion>44.0</em:minVersion>
<em:maxVersion>49.*</em:maxVersion>
```

benbasson:
The slightly sad thing about this is that it would be fine if the current "max" maxVersion was retrieved dynamically from the add-ons site.

kumar303:
> Obviously AMO rejects this.

Can you provide an example XPI that gets rejected by AMO? I didn't think it checks max version anymore because that is a feature of the past. Add-ons are compatible by default in Firefox now IIRC.

benbasson:
> Can you provide an example XPI that gets rejected by AMO? I didn't think it checks max version anymore because that is a feature of the past. Add-ons are compatible by default in Firefox now IIRC.

I could, but I figured the problem was fairly self-explanatory. The minVersion generated is greater than the maxVersion, and AMO throws it out.

kumar303:
Oh, I didn't read closely enough. I see the problem now. We could just set max version to a wild card like `<em:maxVersion>*</em:maxVersion>`. This appears to be supported in my quick test.

asamuzaK:
`"engines": {
    "firefox": ">=44.0 *"
},`
will create:
`<em:minVersion>44.0</em:minVersion>
<em:maxVersion>*</em:maxVersion>`

asamuzaK:
From [jpm/settings.js at master · mozilla-jetpack/jpm](https://github.com/mozilla-jetpack/jpm/blob/master/lib/settings.js):

> // Set a MAX_VERSION to use in install.rdf - this needs to be updated every once in a while
> // see this page for supported ranges: https://addons.mozilla.org/en-US/firefox/pages/appversions/
> exports.MAX_VERSION = "43.0";

Why max version needs to be updated every once in a while?
Why can't "\*" be set by default?
[...]

PR #489, opened 2016-02-09
set max version as a wild card by default

asamuzaK:
JPM should create max version as a wild card `*` by default in install.rdf.
See [jpm xpi creates invalid install.rdf · Issue #461 · mozilla-jetpack/jpm](https://github.com/mozilla-jetpack/jpm/issues/461).


asamuzaK:
[Don't warn about compatibility issues for non-strict add-ons · Issue #1548 · mozilla/addons-server](https://github.com/mozilla/addons-server/issues/1548)

> Thunderbird may support binaries still (Firefox does not)

This pull request may cause regression with AMO.
Closing for now.

PR #490, opened 2016-02-09
watchpost logging

willnwhite:
MDN watchpost documentation and "Developing without browser restarts" sections aren't linked. I spent time working out why extension wasn't logging with watchpost.


kumar303:
Thanks!

Issue #479, opened 2016-01-30
Support localization of add-on's name and description


johannhof:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1218719

Currently you have to manually edit the generated install.rdf file to localize an addon's name or description. It would be cool if JPM could do that automatically.

The easiest way to do it is described in [1218719](https://bugzilla.mozilla.org/show_bug.cgi?id=1218719), where you could just define `NameOfTheAddon_description=localized description of my addon` in `locale/ja.properties`. @kumar303 Does that sound like a good solution?

This is semi-blocking a current work thing of mine, so I'd love to be able to fix it myself soon.

jc3213:
It's a bug reported just by me on bugzilla, I'm now using this way to localize the description(since title is not supported). Here's the sample. or you can get my source from https://github.com/mozilla-jetpack/jpm/files/108789/Simple_Proxy.zip

`var SimplePrefs = require("sdk/simple-prefs");
var Locales = require("sdk/l10n").get;
SimplePrefs.prefs["description"] = Locales("add-on description");`

johannhof:
Yeah it's a good hack but I think having a magic key in the properties file sounds like a good permanent solution (which also works for the add-on title). It should also be easy to implement it in JPM.

jc3213:
@johannhof Anyway,is there any chance that I can use &abc; in package.json to set abc as a l10n key in the locale.properties, as I used to do in options.xul before?

It will be very helpful if there are options that uses the same l10n value.Currently pref-name_options."value in the option label"=translation string is too long. and if more than one option in deferent pref-name has the same string.it will cause lots of bytes to fulfill the l10n as I only use 2 line when it is options.xul

PS:I don't speak English as my native, since I am a Chinese.Maybe https://bugzilla.mozilla.org/show_bug.cgi?id=1210758 has better explanation than I reported here

johannhof:
@jc3213 No worries, I understand what you're looking for. I think it's a reasonable idea, but I'd recommend making a new issue since it's not directly related to the name and description thing :)

asamuzaK:
FYI:
1123428 – Let developers localize add-on metadata with jpm
https://bugzilla.mozilla.org/show_bug.cgi?id=1123428

Provide a way to localize package.json information · Issue #562 · npm/newww
https://github.com/npm/newww/issues/562

johannhof:
[...]

PR #495, opened 2016-02-16
l10n title/description for install.rdf like nodejs

jc3213:
This is based on #419 , But the original one have more than one additional functions and most of them need more tweak to be done. So I've rewritten the l10n part according to jpm code style(maybe?)

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -132,8 +132,8 @@ function createRDF(manifest) {
       var l10n = manifest.locales[locale];
       var l10nDescription = {
         "em:locale": locale,
-        "em:name": l10n.title || manifest.title || manifest.name || "Untitled",
-        "em:description": l10n.description || manifest.description || undefined,
+        "em:name": l10n.title || jetpackMeta["em:name"],
+        "em:description": l10n.description || jetpackMeta["em:description"],
         "em:creator": formatAuthor(manifest.author)
       };
 
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -127,6 +127,24 @@ function createRDF(manifest) {
     });
   }
 
+  if (manifest.locales) {
+    for (var locale in manifest.locales){
+      var l10n = manifest.locales[locale];
+      var l10nDescription = {
+        "em:locale": locale,
+        "em:name": l10n.title || manifest.title || manifest.name || "Untitled",
+        "em:description": l10n.description || manifest.description || undefined,
+        "em:creator": formatAuthor(manifest.author)
+      };
+
+      description.children.push({
+        "em:localized": {
+          "Description": l10nDescription
+        }
+      });
+    }
+  }
+
   var xml = jsontoxml(header, {
     prettyPrint: true,
     xmlHeader: true,
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -130,14 +130,20 @@ function createRDF(manifest) {
   if (manifest.locales) {
     for (var locale in manifest.locales){
       var l10n = manifest.locales[locale];
+      var l10nDescription = {
+        "em:locale": locale,
+        "em:name": l10n.title || manifest.title || manifest.name || "Untitled",
+        "em:description": l10n.description || manifest.description || undefined,
+        "em:creator": formatAuthor(manifest.author)
+      };
+
+      if (manifest.homepage) {
+        l10nDescription["em:homepageURL"] = manifest.homepage;
+      }
+
       description.children.push({
         "em:localized": {
-          "Description": {
-            "em:locale": locale,
-            "em:name": l10n.title || manifest.title || manifest.name || "Untitled",
-            "em:description": l10n.description || manifest.description || undefined,
-            "em:creator": formatAuthor(manifest.author)
-          }
+          "Description": l10nDescription
         }
       });
     }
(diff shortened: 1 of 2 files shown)


jc3213:
Here's a sample to test the functional https://github.com/jc3213/jpm/releases/download/1.0.5/test.zip
Sorry for that, because I don't have a github client

@kumar303 Will you please have a look into this?

johannhof:
Yay, thanks for doing this. :+1: This would fix #479. You should probably add a test if this works properly, though.

jc3213:
@johannhof  That's been done, please have a look into it

jc3213:
I've added original title and description in order to test if only title or description get localized

For exaple: `"locales": "ja": {"title": "ja localed title"}`  or `"locales": "ja": {"description": "ja localed description"}`

lib/rdf.js
```diff
@@ -127,6 +127,28 @@ function createRDF(manifest) {
     });
   }
 
+  if (manifest.locales) {
+    for (var locale in manifest.locales){
+      var l10n = manifest.locales[locale];
+      var l10nDescription = {
+        "em:locale": locale,
+        "em:name": l10n.title || manifest.title || manifest.name || "Untitled",
+        "em:description": l10n.description || manifest.description || undefined,
```
johannhof: nit: `|| undefined` can be removed here

lib/rdf.js
```diff
@@ -127,6 +127,28 @@ function createRDF(manifest) {
     });
   }
 
+  if (manifest.locales) {
+    for (var locale in manifest.locales){
```
johannhof: please add a space between `)` and `{`, your tests are actually failing because of that

johannhof: (you can check the style with `npm run jscs`)

jc3213: oops, sorry for that, it's just a typo, but its creates xpi and signs it correctly so I haven't noticed that

lib/rdf.js
jc3213: I kept that just because there were ` || undefined` in line 52, and line 54 or line 55. I wonder if that 3 can be removed? or I'd just keep it there

johannhof: yeah, no idea what jsontoxml does for other falsy values, so it's probably best to leave it

jc3213:
@johannhof @kumar303 this pull requests couldn't solve `contributors`, `developers`, and `translators` tags, I've done another one but the generated install.rdf has too many CRLF but it supports `contributors`, `developers`, and `translators` tags. See https://github.com/jc3213/jpm/commit/f3e65852c468d4a01544db1f0e0643c38b749ec2

According to document on MDN about package.json https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json It seems there is no need to support `contributors`, no info about `developers` and do supports `translators`. It's really a headache

test/unit/test.rdf.js
```diff
@@ -345,6 +345,102 @@ describe("lib/rdf", function() {
...
+            description: "紹介"
+          },
+          "zh-CN" : {
+            title: "扩展",
+            description: "说明"
+          }
+        }
+      });
+      var locales = xml.getElementsByTagName("em:localized");
+      var locale_ja = locales[0].childNodes[1]; // Description
```
kumar303: all variables typically follow camel case format, such as `localeJa`.

jc3213: @kumar303 That has been done

lib/rdf.js
jc3213: I've changed them to `jetpackMeta["em:name"]` and `jetpackMeta["em:description"]`, and in this way its faster than calculate the name and description again

kumar303:
Thanks for all your work on this and sorry about the headaches

asamuzaK:
Thanks for the fix.
It should be documented too.
[package.json - Mozilla | MDN](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json)

kumar303:
Yes. Could someone edit MDN to add documentation? It's like a wiki so anyone can edit it.

SebastianZ:
If I read the patch correctly, it allows to add the localization directly within the package.json using the `locales` property.
Though it doesn't allow to have the translations within the .properties files as [initially requested](https://github.com/mozilla-jetpack/jpm/issues/479#issue-129976811), i.e. as `<add-on name>.title=Title` and `<add-on name>.description=Description`. Is that correct?
And if so, should I create a new issue for changing that/adding this as a second way?

Sebastian

kumar303:
yes, please create a new issue

SebastianZ:
Thanks for the quick answer! I've created https://github.com/mozilla-jetpack/jpm/issues/527.

Sebastian

SebastianZ:
> Thanks for the fix.
> It should be documented too.
> [package.json - Mozilla | MDN](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json)

Done:

- [package.json - Mozilla | MDN](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json#Key_reference)
- [Localization - Mozilla | MDN](https://developer.mozilla.org/en-US/Add-ons/SDK/Tutorials/l10n#Using_localized_strings_in_add-on_meta_data)

Sebastian

kumar303:
The docs look great! Thank you

PR #498, opened 2016-02-21
Allow relative path for icon url

### LINKED ISSUE #197: Relative icon paths do not work

### LINKED ISSUE #341: Please open issue #197 again. It is not closed!
Here is what works for me in jpm 1.0.1 today with nightly firefox:

    "icon": "resource://issue-pigeon-at-addons-dot-mozilla-dot-org/data/icon48.png",

Having to use this quoted repetition of the id field in jpm 1.0.1 leaves me speechless.

~~Also, icon48.png displays as a 1 mm square.
icon640.png is a size that blends in nicely with the other icons in about:addons (perhaps that's already scaled down in AMO, I have not checked.~~
(I goofed about the size. 48x48px works nicely. I had kept the transparent background size at 640 which made the foreground image show up so small.)

yan12125:
Fixes #197 and fixes #341.

It's a workaround before https://bugzilla.mozilla.org/show_bug.cgi?id=1141839 is resolved. Tested with https://github.com/whuhacker/Unblock-Youku-Firefox on Linux. If this patch is OK I'll try to add some tests.

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -78,6 +78,13 @@ function createRDF(manifest, needsBootstrapJS) {
     delete jetpackMeta["em:icon64URL"];
   }
 
+  /*
+   * When bootstrap.js is not provided by the addon developer, the default
+   * bootstrap.js is used. With this bootstrap.js, the addon's resource:// URI
+   * registration occurs in resource://gre/modules/commonjs/sdk/addon/bootstrap.js.
+   * If a custom bootstrap.js is provided, the program flow mentioned above may
+   * not occur, so the real icon URLs are unknow.
+   **/
   if (needsBootstrapJS) {
     ["em:iconURL", "em:icon64URL"].forEach(function(key) {
       if (jetpackMeta[key] && jetpackMeta[key].indexOf("://") === -1) {
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -10,10 +10,10 @@ var GUIDS = require("./settings").ids;
 var MIN_VERSION = require("./settings").MIN_VERSION;
 var MAX_VERSION = require("./settings").MAX_VERSION;
 
-const UUID_PATTERN = /^\{([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\}$/;
+var UUID_PATTERN = /^\{([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})\}$/;
 
 // Copied from mozilla-central/addon-sdk/source/lib/sdk/bootstrap.js
-function realDomain (id) {
+function realDomain(id) {
   return id.lastIndexOf("@") === 0 ? id.substr(1).toLowerCase() :
     id.toLowerCase().
       replace(/@/g, "-at-").
@@ -79,9 +79,10 @@ function createRDF(manifest, needsBootstrapJS) {
   }
 
   if (needsBootstrapJS) {
-    ['em:iconURL', 'em:icon64URL'].forEach(function (key) {
+    ["em:iconURL", "em:icon64URL"].forEach(function(key) {
       if (jetpackMeta[key] && jetpackMeta[key].indexOf("://") === -1) {
-        jetpackMeta[key] = 'resource://' + realDomain(jetpackMeta["em:id"]) + '/' + jetpackMeta[key];
+        var domain = realDomain(jetpackMeta["em:id"]);
+        jetpackMeta[key] = "resource://" + domain + "/" + jetpackMeta[key];
       }
     });
   }
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -10,6 +10,17 @@ var GUIDS = require("./settings").ids;
 var MIN_VERSION = require("./settings").MIN_VERSION;
 var MAX_VERSION = require("./settings").MAX_VERSION;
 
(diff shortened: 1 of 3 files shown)


johannhof:
For previous discussions around this, see earlier PRs #212 and #267.

kumar303:
@johannhof since you are more familiar, could you review this patch? I think it should be fine since it's not moving files around, right?

johannhof:
@kumar303 sure, I'll take a look! The main issue is that this patch (and the earlier patch) are using `resource:` urls while that protocol might not forever be supported in Firefox add-ons and people would have to repackage with an updated version of jpm. I'm not sure it's a valid argument because 

1) it's _completely broken_ right now which is arguably worse than _will at some point maybe stop to work_ 
2) To fix it in m-c would be equally messy
3) jpm will be obsolete at some point as well

So I would __recommend__ going for this approach, however I'm a bit biased since I wrote the initial patch. Do you know what's up with `resource:`?

kumar303:
Thanks for the detailed info. The fact that it's completely broken as is sounds like a good enough reason to go ahead with this patch.

@yan12125 yes, please continue the patch with tests.

kumar303:
@johannhof also, thanks for your continued help on jpm. I added you as a project member so that you can also help merge patches if you'd like to in the future.

lib/rdf.js
```diff
@@ -64,6 +75,15 @@ function createRDF(manifest) {
     delete jetpackMeta["em:icon64URL"];
   }
 
+  if (needsBootstrapJS) {
```
johannhof: Why would you only do this if no bootstrap.js is present? Is there a reason for that if clause?

johannhof:
@yan12125 in https://github.com/mozilla-jetpack/jpm/pull/267/files#diff-dc54654b4741edd9be2ea246c9563573R61 it logs a warning that using the manifest icon key is not recommended. It would be cool if could include a warning like that.

lib/rdf.js
yan12125: When ```bootstrap.js``` is not provided by the addon developer, [the default ```bootstrap.js```](https://github.com/mozilla-jetpack/jpm-core/blob/master/data/bootstrap.js) is used. With this ```bootstrap.js```, the addon's ```resource://``` URI registration occurs in [resource://gre/modules/commonjs/sdk/addon/bootstrap.js](https://hg.mozilla.org/mozilla-central/file/c59c022943f6/addon-sdk/source/lib/sdk/addon/bootstrap.js#l101). If a custom ```bootstrap.js``` is provided, the program flow mentioned above may not occur, so the real icon URLs are unknow.

yan12125:
> It would be cool if could include a warning like that.

https://developer.mozilla.org/en-US/Add-ons/Install_Manifests says:
```
Starting in Gecko 1.9.2 (Firefox 3.6), you can also simply include your icon, named icon.png, in the base directory of the add-on. This allows your add-on's icon to be displayed even when the add-on is disabled, or if the manifest is missing an iconURL entry.
```
For me the wording _you can also_ indicates both ways are acceptable, and neither is more preferred than another. Are there texts explicit recommending not using ```<em:iconURL>``` or ```<em:iconURL64>``` keys?

For the following statement:
```
This allows your add-on's icon to be displayed even when the add-on is disabled
```
Things may have been changed since this sentence is written. On Firefox 44.0.2, at least EPUBReader, GreaseMonkey, Advanced Cookie Manager and Anonymox uses ```<em:iconURL>``` in ```install.rdf``` and their icons display correctly when the corresponding addon is disabled.

As stated in https://bugzilla.mozilla.org/show_bug.cgi?id=1141839, relative paths should be supported, and with some hacks in jpm it really works. I don't think warnings here are necessary.

Finally, concerning another statement in https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#iconURL:
```
A chrome:// URL to an icon to display in the add-ons list.
```
There should be a note in this section. Something like "relative paths are supported if you're using jpm >= x.y.z" if this PR is finally merged.

yan12125:
@johannhof Can I adapt your test cases from #267 in this PR?

lib/rdf.js
johannhof: Huh, I didn't know this was done in `bootstrap.js`. In any case I suppose we can trust developers using custom `bootstrap.js` with this. Can you please add a very short comment above that line so that people like me understand what's going on? :sweat_smile:

johannhof:
@yan12125:

> Things may have been changed since this sentence is written. On Firefox 44.0.2, at least EPUBReader, GreaseMonkey, Advanced Cookie Manager and Anonymox uses <em:iconURL> in install.rdf and their icons display correctly when the corresponding addon is disabled.

I can still reproduce it on my nightly build using a dummy add-on. Did you reload the page after disabling them? Also, they might still have an `icon.png` in their root.

Apart from that, I would like a warning because falling back to `resource` URLs is still a hack and should be last resort, not the default.

> Can I adapt your test cases from #267 in this PR?

Sure!

johannhof:
@yan12125 do you need any help finalizing this patch? :)

kumar303:
This needs to be rebased on master to resolve the conflicts.

yan12125:
Rebased and tests added.

johannhof:
Ok, thanks! I see you didn't add a warning, which is not a breaker for me. We can make a small follow-up bug for this.

Thank you for contributing. 👍

johannhof:
See #522

PR #499, opened 2016-02-22
upgrade zip-dir version to fix issues on Linux

laurentj:
zip-dir 1.0.1 introduced [a severe issue](https://github.com/jsantell/node-zip-dir/issues/6) where it could use 100% CPU usage. 1.0.2 fixes that.


kumar303:
Thanks, you beat me to it :) I'll cut a release for this shortly.

PR #501, opened 2016-02-22
Fix #481: Don't package signed XPI files of the extension

### LINKED ISSUE #481: JPM should ignore XPIs by default
JPM currently doesn't have `*.xpi` in its default ignore set, which means it will package the last built xpi in the newly built xpi.

freaktechnik:
Currently JPM ignores the XPI files it directly produces if there is no .jpmignore. However the XPI files generated with `jpm sign` have a different naming scheme. This patch also ignores the naming scheme from `jpm sign`. See also #481.

### DIFF
--- a/lib/ignore.js
+++ b/lib/ignore.js
@@ -9,7 +9,6 @@ var when = require("when");
 var Minimatch = require("minimatch").Minimatch;
 var utils  = require("./utils");
 var console = utils.console;
-var getID = require("jetpack-id");
 
 /**
  * Look for .jpmignore in the given directory and use it to filter files
@@ -22,12 +21,9 @@ var getID = require("jetpack-id");
  */
 function ignore(dir, options) {
   var jpmignore = path.join(dir, ".jpmignore");
-  var defaultRules = ["*.zip", ".*", "test/", ".jpmignore"];
+  var defaultRules = ["*.zip", ".*", "test/", ".jpmignore", "*.xpi"];
   return utils.getManifest({addonDir: options.addonDir})
     .then(function(manifest) {
-      // add the xpi file name to the ignore list
-      var xpiName = getID(manifest) + "*.xpi";
-      defaultRules.push(xpiName);
       return fs.exists(jpmignore);
     })
     .then(function(exists) {
--- a/test/unit/test.xpi.js
+++ b/test/unit/test.xpi.js
@@ -231,6 +231,7 @@ describe("lib/xpi", function() {
   it("Test default .jpmignore rules", function(done) {
     process.chdir(extraFilesPath);
     var ID_XPI = "extra-files@jpm.xpi";
+    var ID_SIGNED_XPI = "extra_files-0.0.1-fx.xpi";
     var manifest = require(path.join(extraFilesPath, "package.json"));
 
     // Copy in a XPI since we remove it between tests for cleanup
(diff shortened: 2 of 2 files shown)


kumar303:
It looks like all these test failures are in the integration suite which, admittedly, I find very annoying and hard to debug. I couldn't say what's causing them to fail just by glancing at the test output.

lib/ignore.js
```diff
@@ -28,6 +28,10 @@ function ignore(dir, options) {
       // add the xpi file name to the ignore list
       var xpiName = getID(manifest) + "*.xpi";
       defaultRules.push(xpiName);
+      // add the signed xpi file name to the ignore list
+      var signedXpiName = manifest.title.toLowerCase().replace(/ /g, "_") + "*.xpi";
+      defaultRules.push(signedXpiName);
+      console.log(signedXpiName);
```
kumar303: This console statement isn't necessary. You should be able to see info about the rules when using `--verbose`.

lib/ignore.js
```diff
@@ -28,6 +28,11 @@ function ignore(dir, options) {
       // add the xpi file name to the ignore list
       var xpiName = getID(manifest) + "*.xpi";
       defaultRules.push(xpiName);
+      // add the signed xpi file name to the ignore list
+      var addonTitle = manifest.title || manifest.name || "Untitled";
+      var signedXpiName = addonTitle.toLowerCase().replace(/ /g, "_") + "*.xpi";
```
kumar303: This looks like it will work. If you want to trace the AMO code, this is how it generates filenames https://github.com/mozilla/addons-server/blob/master/src/olympia/files/models.py#L239

lib/ignore.js
```diff
@@ -28,6 +28,11 @@ function ignore(dir, options) {
...
+      defaultRules.push(signedXpiName);
+      console.log(signedXpiName);
```
kumar303: This needs to be removed. You can check the rules with `--verbose`, I think.

kumar303:
Instead of trying to debug these test failures, I'd recommend not touching the integration suite at all for this change. It looks like you can easily add a unit test similar to this one: https://github.com/freaktechnik/jpm/blob/ignore-signed-xpi/test/unit/test.xpi.js#L231

Run them like `npm test unit`.

lib/ignore.js
freaktechnik: Yes, this was left in by accident. The packaged files are indeed listed with `-v`

lib/ignore.js
freaktechnik: Okay, so AMO does actually quite a bit more.

freaktechnik:
Thanks for pointing me at the right test, I didn't spot the ignore tests in the unit tests.

johannhof:
I'm just wondering: why not add `*.xpi` to the default ignore list? Sounds easier and will fix most use cases. Am I missing something?

kumar303:
There might be some legitimate case for including nested XPI files. I just know it's "supported" in AMO because it will check the contents of nested XPIs. I don't know why though. I would agree to a patch that adds `*.xpi` to jpm's default ignore list -- we could just wait until someone complains to see if there is a realistic use case for nested XPIs.

freaktechnik:
From how I understand the code it would mostly be a case of adding a .jpmignore if you want to package up XPIs, since a .jpmignore will cause the default ruleset to be ignored.

johannhof:
Yeah that's why I'd vote +1 for just ignoring `*.xpi`

freaktechnik:
Okay, I've updated the patch to ignore all XPI files by default. Should I squash again or is it fine like this?

kumar303:
Looks good, thanks. Yes, please squash into one commit.

johannhof:
:+1:

PR #502, opened 2016-02-26
more tweak up about localized description

jc3213:
- support `developers`,  `translators`, and `contributors` flag.
- changed the order from `engines` -> `translators` and `contributors` to `translators` and `contributors` -> `engines`
- fixed a bug that if there's no "description" in package.json, it will generate `undefined` in localized description
- some variable changes

### DIFF
--- a/test/unit/test.rdf.js
+++ b/test/unit/test.rdf.js
@@ -352,7 +352,7 @@ describe("lib/rdf", function() {
           "ja" : {
             title: "名前",
             description: "紹介",
-            homepage: "ホームページ"
+            homepage: "ホームページ",
           }
         }
       });
@@ -376,7 +376,7 @@ describe("lib/rdf", function() {
         locales: {
           "ja" : {
             title: "名前",
-            homepage: "ホームページ"
+            homepage: "ホームページ",
           }
         }
       });
@@ -398,7 +398,7 @@ describe("lib/rdf", function() {
         locales: {
           "ja" : {
             title: "名前",
-            description: "紹介"
+            description: "紹介",
           }
         }
       });
@@ -422,7 +422,7 @@ describe("lib/rdf", function() {
         locales: {
           "ja" : {
             description: "紹介",
-            homepage: "ホームページ"
+            homepage: "ホームページ",
           }
         }
       });
@@ -447,12 +447,12 @@ describe("lib/rdf", function() {
           "ja" : {
             title: "名前",
             description: "紹介",
-            homepage: "ホームページ"
+            homepage: "ホームページ",
           },
           "zh-CN" : {
             title: "扩展",
             description: "说明",
-            homepage: "主页"
+            homepage: "主页",
           }
         }
       });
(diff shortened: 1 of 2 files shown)


jc3213:
@johannhof @kumar303  Would you please review this patch?

test/unit/test.rdf.js
```diff
@@ -346,12 +346,13 @@ describe("lib/rdf", function() {
...
-    it("add ja localized title and description to add-on", function() {
-      var xml = setupRDF({ title: "my-title", description: "my-desc",
+    it("add `ja` title, description, and homepage to add-on", function() {
+      var xml = setupRDF({ title: "my-title", description: "my-desc", homepage: "my-page",
         locales: {
           "ja" : {
             title: "名前",
-            description: "紹介"
+            description: "紹介",
+            homepage: "ホームページ"
```
kumar303: this should have a trailing comma

test/unit/test.rdf.js
```diff
@@ -365,13 +366,17 @@ describe("lib/rdf", function() {
...
 
-    it("add ja localized title to add-on", function() {
-      var xml = setupRDF({ title: "my-title", description: "my-desc",
+    it("add `ja` title and homepage to add-on w/o description", function() {
+      var xml = setupRDF({ title: "my-title", homepage: "my-page",
         locales: {
           "ja" : {
-            title: "名前"
+            title: "名前",
+            homepage: "ホームページ"
```
kumar303: another nit: trailing comma

test/unit/test.rdf.js
```diff
@@ -365,13 +366,17 @@ describe("lib/rdf", function() {
       expect(locale.childNodes[3].childNodes[0].data).to.be.equal("名前");
       expect(locale.childNodes[5].tagName).to.be.equal("em:description");
       expect(locale.childNodes[5].childNodes[0].data).to.be.equal("紹介");
+      expect(locale.childNodes[7].tagName).to.be.equal("em:creator");
```
kumar303: I don't see `em:creator` or `author` used in this test. Is it necessary to make an assertion about it?

kumar303:
Thanks for these fixes and features. I think it would be a good idea to finish out the patch (or you could follow up with incremental patches, which is always easier for reviewers).

test/unit/test.rdf.js
jc3213: I don't know, jpm generate `<em:creator/>` by default even you don't have `author` in package.json, this should be the problem with `formatAuthor(manifest.author)`, I think.

You can test it with test2.zip attached in https://github.com/mozilla-jetpack/jpm/pull/502#issuecomment-189050690 by remove the `author` flag, then you'll get
```js
          <em:unpack>false</em:unpack>
          <em:version>0.0.1</em:version>
          <em:name>My-addon</em:name>
          <em:creator/>
          <em:multiprocessCompatible>true</em:multiprocessCompatible>
```

jc3213:
Oops, sorry that I don't quite understand what you've said above. what is incremental patches? sorry, I just learned javascript myself so I'm not very familiar with technique terms

kumar303:
What I meant by incremental patches is to make one pull request (like this one) with a small amount of changes. You can always follow up with more pull requests later. When a pull request has a small amount of changes it is easier to review.

Is this patch ready to merge? It looks good to me but the comment mentions that it might be missing some changes?

jc3213:
@kumar303  Yes, it's ready now. The missing comments were about trailing comma, but I don't know if I should modify `formatAuthor(manifest.author)` part to make it not create `<em:creator/>` tag for install.rdf

test/unit/test.rdf.js
kumar303: I was only talking about the test assertions themselves. But, no worries, more assertions will not be so harmful so we can leave the test as is.

kumar303:
Yes, if you find more problems in the future just open a new pull request.

Thanks for all your work on this patch!

PR #503, opened 2016-02-29
Change default MAX_VERSION to *

wagnerand:
This avoids issues that happen when minversion in install.rdf is greater than MAX_VERSION

@kumar303 r?

### DIFF
--- a/lib/settings.js
+++ b/lib/settings.js
@@ -8,7 +8,7 @@
 exports.MIN_VERSION = "38.0a1";
 // Set a MAX_VERSION to use in install.rdf - this needs to be updated every once in a while
 // see this page for supported ranges: https://addons.mozilla.org/en-US/firefox/pages/appversions/
-exports.MAX_VERSION = "43.0";
+exports.MAX_VERSION = "*";
 // This isn't implemented, so crank up the AOM_SUPPORT_VERSION
 exports.AOM_SUPPORT_VERSION = "50.0a";
 


kumar303:
Can you confirm that this works? Did you try generating an add-on with this and submitting it to AMO?

This was tried before: https://github.com/mozilla-jetpack/jpm/pull/489

My understanding was that you could not submit an add-on to AMO but I could be wrong. There was more discussion about it in: https://github.com/mozilla/addons-server/issues/1548

wagnerand:
Oh, yes, this probably depends on mozilla/addons-server#1548 as we currently allow "*" for webextensions only.

kumar303:
https://github.com/mozilla/addons-server/issues/1548 is just calling for an adjustment to the UI. It isn't calling for a fix to the amo-validator (which I think is what throws the error, right?).

wagnerand:
Yeah, sorry. So @jorgevillalobos are you ok with opening up max-version "*" for at least Firefox Desktop and Android (not sure whether Seamonkey supports jetpack add-ons) for all extensions?

jvillalobos:
Yes, I'm okay with it.

wagnerand:
Filed https://github.com/mozilla/amo-validator/pull/400

wagnerand:
Confirmed that it works now that that amo-validator patch landed.

kumar303:
Great. I will merge it now but I won't release a new jpm until the patch has landed on AMO.

jc3213:
@kumar303  It seems amo already accept * as maximum version number for firefox now. I've signed my addon today with maximum version number *

PR #509, opened 2016-03-05
jpm should not create blank tag <em:creator/>

jc3213:
This is about https://github.com/mozilla-jetpack/jpm/pull/502#discussion-diff-54294423 . I've created a patch previously #508 , but it doesn't changed the npm authors' part, so I create this new one

### DIFF
--- a/lib/rdf.js
+++ b/lib/rdf.js
@@ -255,9 +255,9 @@ function getGUID(type) {
  */
 function formatAuthor(author) {
   if (typeof author === "object") {
-    return (author.name || "").trim();
+    return author.name || undefined;
   }
-  return author || "";
+  return author || undefined;
 }
 
 function createApplication(type, versions) {
--- a/test/unit/test.rdf.js
+++ b/test/unit/test.rdf.js
@@ -220,6 +220,14 @@ describe("lib/rdf", function() {
       var xml = setupRDF({ author: { name: "Marie Curie", email: "mc@espci.fr"} });
       expect(getData(xml, "em:creator")).to.be.equal("Marie Curie");
     });
+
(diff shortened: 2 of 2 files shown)


kumar303:
Can you add a test for this that fails when the values are empty strings (`""`) but passes with your new changes (`undefined`) ?

jc3213:
@kumar303 How about this?
```js
    it("`author` is '', do not ceate <em:creator/>", function() {
      var xml = setupRDF({ author: "" });
      var creator = xml.getElementsByTagName("em:creator");
      expect(creator.length).to.be.equal(0);
    });
```

kumar303:
Yes, that looks like it will work. I think you'll also need an additional test that does `setupRDF({ author: undefined })`

jc3213:
I've add the code following the method of 
```js
    it("multiprocess permission", function() {
      var xml = parseRDF(RDF.createRDF({ id: "1" }));
      expect(nodeExists(xml, "em:multiprocessCompatible")).to.be.equal(false);
      expect(getData(xml, "em:multiprocessCompatible")).to.be.equal(undefined);

      xml = parseRDF(RDF.createRDF({ id: "1", permissions: { multiprocess: true } }));
      expect(nodeExists(xml, "em:multiprocessCompatible")).to.be.equal(true);
      expect(getData(xml, "em:multiprocessCompatible")).to.be.equal("true");

      xml = parseRDF(RDF.createRDF({ id: "1", permissions: { multiprocess: false } }));
      expect(nodeExists(xml, "em:multiprocessCompatible")).to.be.equal(false);
      expect(getData(xml, "em:multiprocessCompatible")).to.be.equal(undefined);
});
```

kumar303:
Thanks for the follow-up, looks great

Issue #512, opened 2016-03-09
jpm sign: support web extensions


kumar303:
Eventually it would be nice if [web-ext](https://github.com/mozilla/web-ext) did signing but it will be quicker to allow JPM to sign web extension XPIs.

kumar303:
web-ext can sign web extensions now, see https://github.com/mozilla/web-ext/issues/145

PR #513, opened 2016-03-09
Sign web extension XPIs

kumar303:
Fixes https://github.com/mozilla-jetpack/jpm/issues/512


kumar303:
This is just a temporary solution because it seemed quicker than adding full signing support to `web-ext`. The real solution should be to extract all the jpm signing logic into a separate module that can be used by both `jpm` and `web-ext`. Until then, this patch will at least allow developers to sign their web extensions.

andymckay:
I'm still r- on this and would recommend that we just move people to web-ext instead. 

It feels like `jpm` is being used by lots of people for the wrong reasons already and the hoops you've have had to go to support signing of `cfx` add-ons and non-jpm add-ons makes me sad. If we do this, we'll be supporting this for ages and having to keep making changes to it and update and that's more bugs. I'd much rather focus our energy on `web-ext`. 

I can see how this might look like a quick win, but it feels like a long term technical debt problem that we'll regret.

kumar303:
Sounds fine to me, I can take the plunge and extract the signing code into a separate module. I'll leave this open for reference until I fully extract everything.

kumar303:
Closed in favor of https://github.com/mozilla/web-ext/issues/145

PR #521, opened 2016-04-03
! Fix error "file already exists" when using command "sign" with XPI file

### LINKED ISSUE #458: jpm sign: Error: EEXIST: file already exists
Hello guys,

I am trying to sign existing xpi file ([firenyx-0.1.14.2014093001.zip](https://github.com/mozilla-jetpack/jpm/files/76180/firenyx-0.1.14.2014093001.zip)) with following command:
```
jpm sign -v --api-key user:613385:367 --api-secret 0318... --xpi firenyx-0.1.14.2014093001.xpi
```

But result is EEXIST error:
```
JPM [info] verbose set
JPM [error] FAIL
Error: EEXIST: file already exists, mkdir 'C:\Users\Martin\AppData\Local\Temp\tmp-2492BkWVhH1SmxAM\content'
    at Error (native)
```

Any idea what I am doing wrong?

node.js version: 4.2.4
jpm version: 1.0.4

PikachuEXE:
Use another unzip library `decompress-zip` to replace `unzip` due to a potential bug in `unzip`

Fixes #458 
Tested command locally

Cause of bug:
A library used `unzip` contains a pontential bug: when unzipping a folder in zip, a file is created instead.
So the file is created multiple times causing the error.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -11,7 +11,7 @@ var tmp = require("tmp");
 var os = require("os");
 var when = require("when");
 var nodefn = require("when/node");
-var unzip = require("unzip");
+var DecompressZip = require("decompress-zip");
 var parse = require("mozilla-toolkit-versioning").parse;
 var compare = require("mozilla-version-comparator");
 var xml2js = require("xml2js");
@@ -204,16 +204,18 @@ function extractXPI(xpiPath) {
       if (err) {
         return cleanUpAndReject(err);
       }
-      fs.createReadStream(xpiPath)
-        .on("error", cleanUpAndReject)
-        .pipe(unzip.Extract({path: tmpPath}))
-        .on("error", cleanUpAndReject)
-        .on("close", function() {
-          resolve({
-            path: tmpPath,
-            remove: removeTmpDir,
-          });
+
+      var unzipper = new DecompressZip(xpiPath);
+      unzipper.on("extract", function(_log) {
+        resolve({
+          path: tmpPath,
+          remove: removeTmpDir,
         });
+      });
+      unzipper.on("error", cleanUpAndReject);
+      unzipper.extract({
+        path: tmpPath,
+      });
     }, {
       prefix: "tmp-extracted-xpi-",
       // This allows us to remove a non-empty tmp dir.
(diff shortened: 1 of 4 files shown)


lib/utils.js
```diff
@@ -204,16 +204,20 @@ function extractXPI(xpiPath) {
...
+      var unzipper = new DecompressZip(xpiPath);
+      unzipper.on("extract", function(_log) {
+        resolve({
+          path: tmpPath,
+          remove: removeTmpDir,
         });
+      });
+      unzipper.on("error", function(err) {
+        cleanUpAndReject(err);
+      });
```
johannhof: you can shorten this to `unzipper.on("error", cleanUpAndReject)`

test/utils/index.js
```diff
@@ -116,9 +116,17 @@ exports.run = run;
...
 function unzipTo(xpiPath, outputDir) {
   return when.promise(function(resolve, reject) {
-    fs.createReadStream(xpiPath)
-      .pipe(unzip.Extract({path: outputDir}))
-      .on("close", resolve);
+    var unzipper = new DecompressZip(xpiPath);
+
+    unzipper.on("extract", function(_log) {
+      resolve();
+    });
```
johannhof: you can shorten this to `unzipper.on("extract", resolve)`

test/utils/index.js
```diff
@@ -116,9 +116,17 @@ exports.run = run;
...
+    unzipper.on("error", function(err) {
+      reject(err);
+    });
```
johannhof: you can shorten this to `unzipper.on("error", reject)`

johannhof:
Hey, thanks for the pull request. I've had some really minor comments, but this LGTM otherwise.

Not sure if we can cover this problem with a test, was there ever a way to reliably reproduce it?

PikachuEXE:
There is no test for passing in an XPI (path), and I don't know how to set it up :S

johannhof:
Ah don't worry about it, if it's a problem with a third-party lib I think we can do without a test. Do you have time to fix the small nits above? :)

PikachuEXE:
Fixed and all green

johannhof:
:+1: Awesome, thanks!

kumar303:
thanks for fixing this!

Issue #531, opened 2016-05-24
Node warning: please update graceful-fs module to a more recent version.


jc3213:
node.exe 6.2.0 (x64) running `jpm -V`, The command-line console alerts `(node:7244) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.` then display `1.0.7`

kumar303:
it's in one of our dependencies. We just need to track it down and upgrade the dependency (if it has been fixed).

freaktechnik:
I think npm install actually says where the deprecated graceful-fs version is coming from.

jc3213:
I was using jpm xpi to package xpi files and the console warns about it `(node:34596) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.`

And when using npm install jpm, many module has been alterted that is deprecated and I can hardly remember one of them alerts that the old version will not be compatible with nodejs 7.0.0

Croydon:
`npm ls graceful-fs -g` does the magic.

[fs-extra](https://www.npmjs.com/package/fs-extra) needs to get updated (directly in this repo).

PR #535, opened 2016-06-07
Update fs extra dependency

asamuzaK:
Fix [ #531](https://github.com/mozilla-jetpack/jpm/issues/531)


kumar303:
Thanks! I'll merge it in. It looks like mocha also needs updating to make the warning go away.

PR #537, opened 2016-06-10
Add test on node.js 6

asamuzaK:
Adding node.js version 6 to travis CI
BTW is it still necessary to test on 0.10?


johannhof:
Thanks! As long as it's not failing I'd keep it, there are likely still users in e.g. corporate environments or with legacy requirements that need 0.10 support.

Issue #540, opened 2016-06-17
Fail if there's a manifest.json


andymckay:
If the package has a `manifest.json` in the root (it's a WebExtension) then we should probably just fail and raise a nice error message saying "please go use web-ext". This will help people who hit issues like #539

rpl:
This issue seems to be possible only if the add-on directory contains both a `manifest.json` and a `package.json` file (which can be the case if a WebExtension is built from a directory which is also an npm package).

kumar303:
fixed in https://github.com/mozilla-jetpack/jpm/pull/541

PR #541, opened 2016-06-20
Raise an error when jpm is used against a WebExtension add-on

rpl:
This PR is related to #540 (and depends from jsantell/jetpack-validation#4)

- [x] land fix in jetpack-validation
- [x] wait for the new jetpack-validation version (0.0.6) to be published on npm
- [x] update this pull request to require the new jetpack-validation 0.0.6 in the package dependencies


kumar303:
I pushed up the change for you https://www.npmjs.com/package/jetpack-validation

kumar303:
(we also moved its repo to the `mozilla-jetpack` project)

rpl:
@kumar303 I've updated this pull request to use the change merged in the jetpack-validation dependency, and the travis build is now green as expected.

test/unit/test.xpi.js
```diff
@@ -509,4 +510,21 @@ describe("lib/xpi", function() {
...
   });
+
+  it("Warn when executed against a WebExtension", function(done) {
+    process.chdir(webextensionUnsupportedPath);
+
+    // In this test scenario, jpm is executed agains a WebExtension that contains
+    // both a package.json and a manifest.json files in the root dir, probably because
+    // the add-on is being built from an npm package.
+    var manifest = require(path.join(webextensionUnsupportedPath, "package.json"));
+    xpi(manifest).then(function () {
```
kumar303: it would be better to return this promise object (and remove `done` from the function arguments) so that throwing an exception here will fail the promise chain and cause a real test failure.

kumar303: oh, I see you are catching at the end anyway so it will work as is.

kumar303: This is fine as is. It is consistent with the style already in this test module :)

kumar303:
Looks good!

Issue #315, opened 2015-08-05
jpm xpi needs option to specify the output


bgirard:
Hit this while migrating from cfx and this is breaking my build script because now the output file now changes each version.

lepolt:
+1, this breaks our build scripts (which try to be as dumb and generic as possible).

Standard8:
@kumar303 I believe this is fixed now with commit 8059d86?

kumar303:
Ah, yes it was. See https://github.com/mozilla-jetpack/jpm/pull/544 for discussion.

PR #544, opened 2016-06-29
Fixing #315 : command line option to specify the output directory for the xpi package

alexduch:
Hi  @kumar303, @gozala, @jsantell, @Mossop 

Here is a simple change that fixes issue #315 . Could one of you review it ?
Please let me know if this change is not relevant, or if I need to provide anything else.

Thank you very much in advance.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -74,8 +74,7 @@ program
           // turn into an absolute path
           function(dir) {
             return path.resolve(dir);
-          },
-          process.cwd())
+          })
   .action(cmd.prepare("xpi", program, function(manifest, options) {
     var start = Date.now();
     xpi(manifest, assign({}, program, options)).then(function(xpiPath) {
--- a/bin/jpm
+++ b/bin/jpm
@@ -69,6 +69,13 @@ program
 program
   .command("xpi")
   .description("Bundle the addon into an .xpi file")
+  .option("--dest-dir <output_dir>", "Where to write the generated xpi",
+          // turn into an absolute path
+          function(dir) {
+            program.xpiPath = path.resolve(dir);
+            return program.xpiPath;
+          },
+          process.cwd())
   .action(cmd.prepare("xpi", program, function(manifest) {
     var start = Date.now();
     xpi(manifest, program).then(function(xpiPath) {
--- a/bin/jpm
+++ b/bin/jpm
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 var VERSION = require("../package.json").version;
+var assign = require("object-assign");
 var program = require("commander");
 var console = require("../lib/utils").console;
 var run = require("../lib/run");
@@ -72,13 +73,12 @@ program
   .option("--dest-dir <output_dir>", "Where to write the generated xpi",
           // turn into an absolute path
           function(dir) {
-            program.xpiPath = path.resolve(dir);
-            return program.xpiPath;
+            return path.resolve(dir);
           },
           process.cwd())
-  .action(cmd.prepare("xpi", program, function(manifest) {
+  .action(cmd.prepare("xpi", program, function(manifest, options) {
     var start = Date.now();
-    xpi(manifest, program).then(function(xpiPath) {
+    xpi(manifest, assign({}, program, options)).then(function(xpiPath) {
       var diff = Date.now() - start;
       console.log("Successfully created XPI at " +
                   xpiPath + " (" + diff + "ms)");
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -49,8 +49,9 @@ function xpi(manifest, options) {
   }
   xpiName += ".xpi";
   var updateRdfName = getID(manifest) + xpiVersion + ".update.rdf";
-  var xpiPath = join(options.xpiPath || options.destDir || addonDir, xpiName);
-  var updateRdfPath = join(options.xpiPath || addonDir, updateRdfName);
+  var outputPath = options.xpiPath || options.destDir || addonDir;
+  var xpiPath = join(outputPath, xpiName);
+  var updateRdfPath = join(outputPath, updateRdfName);
   var useFallbacks = !(options.forceAOM || hasAOMSupport(manifest));
 
   options = _.merge(options, {
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -20,8 +20,9 @@ var _ = require("lodash");
  *
  * @param {Object} manifest
  * @param {Object} options
- *   - `xpiPath` Path where xpi should be saved. Used by `run`, `test` to indicate
- *               that the xpi saved in a temp directory.
+ *   - `destDir` Directory path where xpi should be saved.
+ *   - `xpiPath` Legacy alias for `destDir`. Used by `run` and `test` to
+ *               save the xpi to a temp directory.
  * @return {Promise}
  */
 
@@ -48,7 +49,7 @@ function xpi(manifest, options) {
   }
   xpiName += ".xpi";
   var updateRdfName = getID(manifest) + xpiVersion + ".update.rdf";
-  var xpiPath = join(options.xpiPath || addonDir, xpiName);
+  var xpiPath = join(options.xpiPath || options.destDir || addonDir, xpiName);
   var updateRdfPath = join(options.xpiPath || addonDir, updateRdfName);
   var useFallbacks = !(options.forceAOM || hasAOMSupport(manifest));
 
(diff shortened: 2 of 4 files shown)


bin/jpm
```diff
@@ -69,6 +69,12 @@ program
 program
   .command("xpi")
   .description("Bundle the addon into an .xpi file")
+  .option("--xpiPath <output_dir>", "Where to write the generated xpi",
```
kumar303: For consistency, this should be all lowercase. Also, I would recommend calling this `--dest-dir` to make it a bit more clear that it's a destination directory, not the actual XPI file path.

kumar303: oh, I see what you were doing. Hmm, `options.xpiPath` was a poor choice of names to indicate a destination directory :) Check the [commander docs](https://github.com/tj/commander.js) to see what the best way is to turn the `--dest-dir` value into `options.xpiPath`.

freaktechnik:
The failures here seem to be the same as in #546

bin/jpm
```diff
@@ -69,6 +69,13 @@ program
 program
   .command("xpi")
   .description("Bundle the addon into an .xpi file")
+  .option("--dest-dir <output_dir>", "Where to write the generated xpi",
+          // turn into an absolute path
+          function(dir) {
+            program.xpiPath = path.resolve(dir);
+            return program.xpiPath;
```
kumar303: This looks fragile to me because it relies on a global variable (for example, a refactoring might rename `program` and miss this part). It would be more direct like:

    return path.resolve(dir);

Also, what happens when `--dest-dir` is not given on the command line?

I think instead of making this set `program.xpiPath` that you should modify the signature of `xpi()` so that it looks for the `program.destDir` option.

kumar303:
You can rebase on master to get a fix for the failing Firefox tests

kumar303:
Have you had a chance to try rebasing this on master and addressing the issues? Let me know if you have questions.

alexduch:
Hi !

Sorry, I didn't have much time to work on this lately.
Thanks for having fixed the tests.
I have rebased the changes, I will take a closer look during the next days to the issues you raised, and to my commit messages that seem to break some rules.

Best regards.

kumar303:
Thanks! The tests are passing now, it's just nagging you about formatting the commit message. You will probably want to squash all commits as well. https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

alexduch:
Regarding the behaviour when `--dest-dir` is not given on the command line, the default value `process.cwd()` is used.

Otherwise, I don't really feel like modifying `xpi()` because it might impact other parts of the code, and I wanted this change to be as simple as possible.
Additionally, when defining the option for the 'xpi' command only, `program.destDir` is not set, `destDir` is added to a different `Command` object returned by `program.command("xpi")`.

I see 2 possibilities to keep this change simple and not impact other parts of the code:
1) Keep assigning `program.xpiPath` in the option callback.
2) Assign it in the callback passed to `cmd.prepare()` using the commanderOptions parameter.

Another possibility would be to define the option on the `program` object, but the impact on the other commands has to be checked, and we would still need to assign `program.xpiPath` from `program.destDir` somewhere.

kumar303:
Relying on the `program.xpiPath` assignment still feels like too much of a fragile side effect. I know the code base already has confusing magic like this but I'd rather not introduce more.

What you discovered about the two different options objects can be fixed by adjusting the callback. What do you think about this approach below? I understand that fixing all existing calls to use the new `destDir` option could get hairy -- that can be skipped.

It would be great to see a unit test added to [test.xpi.js](https://github.com/mozilla-jetpack/jpm/blob/master/test/unit/test.xpi.js) to cover `destDir` if it's not too tricky. I wouldn't bother with the functional tests because they are a pain to run.

````diff
diff --git a/bin/jpm b/bin/jpm
index 03ce13a..e9fffa1 100755
--- a/bin/jpm
+++ b/bin/jpm
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 var VERSION = require("../package.json").version;
+var assign = require("object-assign");
 var program = require("commander");
 var console = require("../lib/utils").console;
 var run = require("../lib/run");
@@ -72,13 +73,12 @@ program
   .option("--dest-dir <output_dir>", "Where to write the generated xpi",
           // turn into an absolute path
           function(dir) {
-            program.xpiPath = path.resolve(dir);
-            return program.xpiPath;
+            return path.resolve(dir);
           },
           process.cwd())
-  .action(cmd.prepare("xpi", program, function(manifest) {
+  .action(cmd.prepare("xpi", program, function(manifest, options) {
     var start = Date.now();
-    xpi(manifest, program).then(function(xpiPath) {
+    xpi(manifest, assign({}, program, options)).then(function(xpiPath) {
       var diff = Date.now() - start;
       console.log("Successfully created XPI at " +
                   xpiPath + " (" + diff + "ms)");
diff --git a/lib/xpi.js b/lib/xpi.js
index 3f29e1f..36faaff 100644
--- a/lib/xpi.js
+++ b/lib/xpi.js
@@ -20,8 +20,9 @@ var _ = require("lodash");
  *
  * @param {Object} manifest
  * @param {Object} options
- *   - `xpiPath` Path where xpi should be saved. Used by `run`, `test` to indicate
- *               that the xpi saved in a temp directory.
+ *   - `destDir` Directory path where xpi should be saved.
+ *   - `xpiPath` Legacy alias for `destDir`. Used by `run` and `test` to
+ *               save the xpi to a temp directory.
  * @return {Promise}
  */
 
@@ -48,7 +49,7 @@ function xpi(manifest, options) {
   }
   xpiName += ".xpi";
   var updateRdfName = getID(manifest) + xpiVersion + ".update.rdf";
-  var xpiPath = join(options.xpiPath || addonDir, xpiName);
+  var xpiPath = join(options.xpiPath || options.destDir || addonDir, xpiName);
   var updateRdfPath = join(options.xpiPath || addonDir, updateRdfName);
   var useFallbacks = !(options.forceAOM || hasAOMSupport(manifest));
 

````

**Comment by kumar303** [comment]
Let me know if you have questions about my last comment. Thanks for working on this!

File: test/unit/test.xpi.js [hunk 3]
```diff
@@ -52,6 +53,26 @@ describe("lib/xpi", function() {
     })
     .catch(done);
   });
+
+  it("Zips up cwd's addon in the specified directory", function(done) {
```
kumar303: if you remove `done` from the argument list then you can simply return a promise. You won't need to call `done()` and you won't need to add a `catch()` handler.

test/unit/test.xpi.js
```diff
@@ -52,6 +53,26 @@ describe("lib/xpi", function() {
...
+    process.chdir(simpleAddonPath);
+    var manifest = require(path.join(simpleAddonPath, "package.json"));
+    var xpiPath;
```
kumar303: looks like this doesn't need to be declared in this scope

kumar303:
It's looking good, thanks. I just added some minor clean-up comments.

I'm not sure why the tests are failing. I see "JavaScript error: chrome://mozapps/content/extensions/extensions.js, line 769: NS_ERROR_FAILURE: Invalid view: detail" but I'm not sure if that's meaningful.

freaktechnik:
Hmm, this seems to break one test, not sure why it fails, but it looks like an add-on file is not generated where it is expected?

bin/jpm
```diff
@@ -69,9 +70,15 @@ program
 program
   .command("xpi")
   .description("Bundle the addon into an .xpi file")
-  .action(cmd.prepare("xpi", program, function(manifest) {
+  .option("--dest-dir <output_dir>", "Where to write the generated xpi",
+          // turn into an absolute path
+          function(dir) {
+            return path.resolve(dir);
+          },
+          process.cwd())
```
Standard8: The problem with the test failing is here. The `process.cwd()` changes the default location for the xpi when `--dest-dir` is not specified.

Currently the test effectively invokes jpm with `--addon-dir` set to a path, but not setting the `--dest-dir`. Hence the output directory is now set to `process.cwd()` whereas before it would have been the same location as `--addon-dir`.

I'd suggest just dropping the `process.cwd()` argument from this completely, and then when `--dest-dir` isn't specified we'll get the same behavior as jpm currently has.

lib/xpi.js
```diff
@@ -48,7 +49,7 @@ function xpi(manifest, options) {
   }
   xpiName += ".xpi";
   var updateRdfName = getID(manifest) + xpiVersion + ".update.rdf";
-  var xpiPath = join(options.xpiPath || addonDir, xpiName);
+  var xpiPath = join(options.xpiPath || options.destDir || addonDir, xpiName);
   var updateRdfPath = join(options.xpiPath || addonDir, updateRdfName);
```
Standard8: Since `update.rdf` is also an output file, could we change that location as well please?

alexduch:
I can't figure out why this test is failing for the moment.
I don't know if the add-on file is generated elsewhere, or not generated at all.

alexduch:
Thanks for the help !

kumar303:
I gave it spin and it's working well. Thanks again for all the follow-ups.

PR #547, opened 2016-07-12
Add support for the multiprocessCompatible flag

johannhof:
See https://developer.mozilla.org/en-US/Add-ons/Install_Manifests#multiprocessCompatible

Add-On SDK developers should have an easy way to set this without
manually touching the install.rdf. This patch allows developers to add
the flag to their package.json files.


diox:
This already exists in `jpm`. Have a look at https://github.com/johannhof/jpm/blob/3c2574b1c9349623f138c2eff8eb3dccbcb7567a/lib/rdf.js#L104 ; MDN references this key in various pages, such as https://developer.mozilla.org/en-US/Add-ons/Working_with_multiprocess_Firefox

johannhof:
Duh, thanks.

Issue #468, opened 2016-01-23
jpm run: do temporary extension install


freaktechnik:
It seems that currently JPM isn't installing extensions as "temporary" yet, when supported. From Firefox 45 and onwards unsigned extensions can be temporarily installed (for example with about:debugging). Even though the extension ran with jpm run appears in about:debugging, it still gets disabled when resetting the `xpinstall.signatures.required` pref back to true.

In order for JPM to continue working with the Firefox 46 Beta or release and later.

kumar303:
How would this be helpful for development with jpm? It already installs an add-on in a temporary profile which achieves the same effect.

Anyway, I'd be more interested in adding new features like this to [web-ext](https://github.com/mozilla/web-ext) rather than jpm.

freaktechnik:
So you could still use release builds, since from what I understand about:debugging will also be available on release builds.

I understand that for you this is more interesting to implement in web-ext, however jetpack is at least still supported until XPCOM is artificially deprecated from what I understand.

kumar303:
Yes, `jpm` is still supported so I will leave this issue open for it. I would like to minimize the amount of duplicated effort between web-ext and jpm though, if possible!

As for the use case of testing extensions in release builds, I can see how this feature will be useful.

freaktechnik:
I guess it would be best if fx-runner could do this, though you might not want to use that in the web-ext tool, but that's not up to me.

kumar303:
Yes, I am using `fx-runner` in web-ext currently but so far I have had to replace its `addExtension()` function -- it doesn't support web extensions. My plan is to use as much of it as I can so if someone patches it for temporary extensions then that would be nice :)

kumar303:
> I have had to replace its addExtension() function 

I might try to uplift this patch. It's still in the "bang on everything until it works" stage.

freaktechnik:
This will become an issue from the 2nd of August on, since then Release and Beta will both no longer have the pref. This mean the travis tests will break until either this is fixed or the used Firefox build is changed.

kumar303:
does temporary installation even work on Release and Beta? Doesn't the signature restriction still apply?

kumar303:
It may be better to use [unbranded builds](https://wiki.mozilla.org/Addons/Extension_Signing#Unbranded_Builds) for jpm development.

freaktechnik:
about:debugging works on release and beta.

kumar303:
[...]

PR #561, opened 2016-07-30
fix: Avoid JS errors from external file from normandy.cdn.mozilla.org

freaktechnik:
When running tests I'd see JS errors from a selfrepair.js. This change makes Firefox not download that file.

### DIFF
--- a/data/preferences.js
+++ b/data/preferences.js
@@ -67,7 +67,10 @@ exports.DEFAULT_FIREFOX_PREFS = {
     "browser.safebrowsing.provider.0.updateURL" : "http://localhost/safebrowsing-dummy/update",
 
     // Disable self repair/SHIELD
-    "browser.selfsupport.url": "https://localhost/selfrepair"
+    "browser.selfsupport.url": "https://localhost/selfrepair",
+
+    // Disable Reader Mode UI tour
+    "browser.reader.detectedFirstArticle": true
 };
 
 // When launching a temporary new Thunderbird profile, use these preferences.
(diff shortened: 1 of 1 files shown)


freaktechnik:
Hmm, `"browser.reader.detectedFirstArticle": true` would disable the reader view UI tour at the startup. Should I add that to this patch or make a separate pull?

kumar303:
> would disable the reader view UI tour

I'd say go ahead and add that pref to this patch

kumar303:
Thanks!

kumar303:
@freaktechnik thanks again for this. If you feel like it, you could also port this to web-ext :) https://github.com/mozilla/web-ext/issues/406 but it's not urgent

asamuzaK:
Looks like Travis is failing... [Build #780 - mozilla-jetpack/jpm - Travis CI](https://travis-ci.org/mozilla-jetpack/jpm/builds/149329466)

freaktechnik:
This is likely due to travis using the release Firefox version and that one no longer working with jpm. (See #468 )

kumar303:
Hmm, I don't see [explicit support](https://docs.travis-ci.com/user/firefox/) from TravisCI for nightly or dev-edition. ESR gets updated infrequently so it may not be a good candidate.

We used to have code that [manually downloaded and installed a firefox binary](https://github.com/mozilla-jetpack/jpm/blob/master/.travis.yml#L9-L13) so it shouldn't be too hard to resurrect that and point it at an [unbranded build](https://wiki.mozilla.org/Addons/Extension_Signing#Unbranded_Builds).

PR #566, opened 2016-08-10
fix: exclude the manifest.json from the xpi

### LINKED ISSUE #565: Do not throw error when both a manifest.json and package.json file exist
In JPM [1.1.3](https://github.com/mozilla-jetpack/jpm/releases/tag/1.1.3) we released https://github.com/mozilla-jetpack/jpm/commit/28e803d3 which throws an error if you mix `package.json` and `manifest.json`:

````
JPM [error] XPI creation failed: Error: Both "package.json" and "manifest.json" files found in the add-on root dir.

If the add-on is a WebExtension, you should use web-ext instead of jpm for building and signing your add-on XPI file:

- https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Getting_started_with_web-ext
````

I think this was the wrong decision because one development use case is to mix SDK add-on source code with Chrome extension source code. In these cases, they may not be able to fully port to a Firefox WebExtension (e.g. for older Firefox support or simply to save time).

We should turn this into a warning instead of a hard error.

See https://github.com/mozilla-jetpack/jpm/issues/564

### PR DESCRIPTION
**Pull request by rpl**
This change turns the presence of a "manifest.json" file in an Addon SDK add-on source dir from an error into a warning and always exclude the "manifest.json" from the built xpi (because the addons server will not accept the submission of an xpi if both the "manifest.json" and the "install.rdf" are included in an add-on).

This PR is intended as the proposed change to fix #565.

### CODE REVIEW

**Comment by rpl** [comment]
Unfortunately the produced warning message is not tested anymore because it is now printed directly in the console instead of raising a catchable exception :-(

**Comment by kumar303** [comment]
The changelog linter doesn't like the length of your message but I don't think that's such a useful rule. I suggest turning that check into a warning [like I did on web-ext](https://github.com/mozilla/web-ext/blob/master/.conventional-changelog-lintrc#L5).

File: lib/ignore.js [hunk 1]
```diff
@@ -54,6 +57,8 @@ function ignore(dir, options) {
       }
     })
     .then(function(lines) {
+      // Add "manifest.json" to always exclude it from the xpi.
+      lines = ["manifest.json"].concat(lines);
```
kumar303: does it still need to be in the `defaultRules` if it's always added here?

test/unit/test.xpi.js
```diff
@@ -511,20 +516,66 @@ describe("lib/xpi", function() {
...
-      expect(error.message).to.contains("Getting_started_with_web-ext");
-      done();
-    }).catch(done);
+    var manifest = require(path.join(webextensionManifestExcludedPath, "package.json"));
+    return xpi(manifest)
+      .then(function(xpiPath) {
+        return utils.unzipTo(xpiPath, tmpOutputDir);
+      })
+      .then(function(xpiPath) {
+        var testDoesNotExist = when
```
kumar303: can't this just be `return when.all(["manifest.json"].map(...))` ? Why is assigning it to a var necessary?

test/unit/test.xpi.js
```diff
@@ -511,20 +516,66 @@ describe("lib/xpi", function() {
...
+          .all([ "manifest.json" ]
+          .map(function(p) { return path.join(tmpOutputDir, p); })
```
kumar303: this block would be easier for me to read if it was indented to show how it's a method on the Array instance.

test/unit/test.xpi.js
```diff
@@ -511,20 +516,66 @@ describe("lib/xpi", function() {
...
+          .map(function(p) { return fs.exists(p); }))
+          .then(function(results) {
+            results.forEach(function(exists) {
+              expect(exists).to.be.equal(false);
+            });
+          });
+
+        return when.all([ testDoesNotExist ]);
+      })
```
kumar303: If you remove `done` from the function argument, you can end the promise chain here. If you return a promise, you wouldn't need to make assertions about the catch or call `don()` in the `then` handler.

kumar303: probably just copy and paste of some legacy stuff

test/unit/test.xpi.js
```diff
@@ -511,20 +516,66 @@ describe("lib/xpi", function() {
...
+      .catch(function(error) {
+        // No errors should received here
+        expect(error).to.be.equal(null);
+      })
+      .then(function () { done(); }, done);
+  });
+
+    it("Exclude a manifest.json even if not included in a .jpmignore file", function(done) {
```
kumar303: indented too far

kumar303:
Looking good, mostly just some nits. Sorry you had to battle with the function suite :/ If you didn't include a functional test I probably would not have complained :)

rpl:
@kumar303 Thanks for #568! jpm tests are now able to complete on travis :tada: 

I've pushed an updated version of the patch that exclude the manifest.json file from the xpi (with all the suggested changes applied), as well as a new patch that relaxes the changelog linting test as suggested.

kumar303:
Thanks!

PR #570, opened 2016-08-12
test: Use get-firefox to load latest unbranded release build

### LINKED ISSUE #562: Fix travis builds not being able to install unsigned extensions
There are four solutions to this:

 - Use the latest ESR (I think this will work until the next major ESR version)
 - Specify that we want to test against Firefox 47.0
 - Use a tool to download an unbranded build or dev-edition/nightly (dev-edition and nightly are easier afaik)
 - Fix #468 

Thanks @asamuzaK for pointing the build failures of the merge of #561 out.

freaktechnik:
Switch to using a release equivalent build for running tests instead of nightly. Fixes #562 for good.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 sudo: false
 language: node_js
 node_js:
-- '0.10'
 - '4'
 - '5'
 - '6'
@@ -25,10 +24,8 @@ before_install:
 install:
 - npm install
 before_script:
-# Install Firefox Nightly to prevent signing errors.
-- cd $TRAVIS_BUILD_DIR/../
-- wget "https://download.mozilla.org/?product=firefox-nightly-latest&os=linux64&lang=en-US" -O firefox.tar.bz2 && tar xvf firefox.tar.bz2
-- cd $TRAVIS_BUILD_DIR
+# Install unbranded Firefox to prevent signing errors.
+- npm run get-unbranded-firefox -- --extract-to ../
 script:
 - export JPM_FIREFOX_BINARY=$TRAVIS_BUILD_DIR/../firefox/firefox-bin
 - npm run-script jshint
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
     "test": "node ./test/run.js",
     "prepush": "npm run jshint && npm run jscs && npm run documentation",
     "changelog": "conventional-changelog -p angular -u",
-    "changelog-lint": "conventional-changelog-lint --from master"
+    "changelog-lint": "conventional-changelog-lint --from master",
+    "get-unbranded-firefox": "get-firefox -ecb unbranded-release"
   },
   "homepage": "https://github.com/mozilla-jetpack/jpm",
   "repository": {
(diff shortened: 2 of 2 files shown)


freaktechnik:
Right, so get-firefox needs node >=4.0.0, so either only node 4 and bigger get unbranded builds, or I make it work for node 0.10, which I don't really want to.

.travis.yml
```diff
@@ -26,8 +26,9 @@ install:
 - npm install
 before_script:
 # Install Firefox Nightly to prevent signing errors.
+- npm i -g get-firefox
```
kumar303: can't this just go into package.json under dev requirements?

freaktechnik: It can yes. Would you prefer to then call the binary from the `.travis.yml` or with an npm script?

kumar303:
You can remove 0.10 from `.travis.yml`. Node 4 has been the [active LTS](https://github.com/nodejs/LTS) for a couple months now.

.travis.yml
kumar303: if you call it from an npm script then you won't have to worry about adjusting the PATH

freaktechnik: No need to adjust the path, I'd just have to call it from `node_modules/.bin/get-firefox`, however it is weird to have it as dep then.

freaktechnik:
Okay, so for this to work I first have to teach get-firefox to not only extract in-place (should be quite simple).

freaktechnik:
Hm, I'm not sure what's wrong now.

freaktechnik:
whew, I screwed that up.

freaktechnik:
\o/

kumar303:
Huh. Do you have any idea why Travis shows `get-firefox` taking 20-30 seconds? Downloading Nightly only takes about 8 seconds (you can see it in one of the master builds). Maybe Nightly is a smaller download?

kumar303:
This is great! Thanks for setting this up.

freaktechnik:
I think it's mostly due to the way I download it. Instead of streaming it, it gets downloaded with node-fetch and then written to disk. I should look into skipping the writing to disk part and directly extract that buffer.

Seapking of which, this puts the .tar.bz2 into the source root dir, but I guess that gets cleaned up before it would be deployed?

kumar303:
> I guess that gets cleaned up before it would be deployed?

I don't think we need to worry about it. The deployment is just `npm publish` from the root directory which will only upload the contents of the package, nothing outside of that directory.

freaktechnik:
The .tar.bz2 will get downloaded to the package root directory^, since using npm scripts requires the cwd to be where the package.json is.

kumar303:
oh, I see. Hmm, I don't really know what `npm publish` does -- would it distribute all files in the package root? Hopefully not.

freaktechnik:
I'm working on a get-firefox version that will directly use the buffer of the downloaded file instead of writing the archive when extracting it, so that should solve this. (Mostly figuring out how the checksum checks will work)

PR #572, opened 2016-08-15
test: Update get-firefox to version that doesn't write archive to disk

freaktechnik:
This version of `get-firefox` addresses writing the archive to disk and instead directly unzips it to the target directory.


kumar303:
good idea, thanks!

PR #574, opened 2016-08-25
Fix #329: Expose --xpi-path to the command line, so the output directory for the xpi can be set.

### LINKED ISSUE #329: Unable to set xpiPath option when creating XPI
I'm trying to create an XPI and save it to a specific directory, but am having issues.

None of these seem to currently be working:

```sh
$ ./node_modules/.bin/jpm xpi --xpiPath=dist/
  error: unknown option `--xpiPath`

$ ./node_modules/.bin/jpm xpi --xpi-path=dist/
  error: unknown option `--xpi-path`
```

Standard8:
As discussed in #329 it would be nice to be able to set the output directory for the xpi.

This just exposes the already existing option `xpiPath` to the command line as `--xpi-path`.

There's no default supplied so the existing behavior without `--xpi-path` will be unaffected.

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -41,6 +41,11 @@ program
             return path.resolve(dir);
           },
           process.cwd())
+  .option("--xpi-path <path>", "Output directory path for the add-on",
+          // turn into an absolute path
+          function(dir) {
+            return path.resolve(dir);
+          })
   .option("--binary-args <CMDARGS>", "Pass additional arguments into Firefox.")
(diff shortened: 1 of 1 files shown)


kumar303:
I've already left some detailed feedback in https://github.com/mozilla-jetpack/jpm/pull/544 so I'd rather continue with that patch. However, if the contributor is not able to finish it out then we could start again with a new patch.

kumar303:
Thanks for starting on this patch! We just landed https://github.com/mozilla-jetpack/jpm/pull/544 so I think that covers what you needed.

PR #577, opened 2016-09-01
using `--profile` doesn't re-use prefs, for complex reasons.

gregglind:
Claim:
1.  expect --profile (copy or otherwise) to start with same user.js
2.  MAYBE the jpm prefs should override?  I dunno

What happens:
1.  We end up with the `default` set from `firefox-profile` package, which is not what we want.

Why:

- `new FirefoxProfile(aPath)` creates a profile not using the existing prefs.

I am open to ideas about what the 'right' behaviour should be.

Easy fix, once we decide.

(Fwiw, `firefox-profile` 0.4.0, the next in the series, changes the new FirefoxProfile interface and is ALSO buggy / wrong when cloning.  BE WARNED)

### DIFF
--- a/lib/profile.js
+++ b/lib/profile.js
@@ -2,15 +2,41 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 "use strict";
-
+var fs = require("fs");
 var FirefoxProfile = require("firefox-profile");
 var FirefoxProfileFinder = require("firefox-profile/lib/profile_finder");
 var getPrefs = require("./preferences").getPrefs;
 var when = require("when");
 var console = require("./utils").console;
 
+// revised from firefox_profile, b/c node fs module required encoding
+function readExistingUserjs(profile) {
+  var self = profile;
+  var regExp = /user_pref\('(.*)',\s(.*)\)/;
+  var contentLines = fs.readFileSync(self.userPrefs, {encoding: "utf8"}).
+    split("\n");
+  contentLines.forEach(function(line) {
+    var found = line.match(regExp);
+    if (found && found[1]) {
+      self.defaultPreferences[found[1]] = found[2];
+    }
+  });
+}
+
+function setAppPrefs(profile) {
+  // Set default preferences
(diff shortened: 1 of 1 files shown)


kumar303:
Can you explain your use case here? If you're already copying an existing profile, why don't you just put the ones from `user.js` into the custom profile before you copy it?

gregglind:
## Use case:

- for `Shield Studies`, we need to do QA testing for some obscure things:  

   - making sure orientation only happens in certain branches, and only during `install` or `uninstall`
   - setting a bunch of state before startup (`--prefs` works, but could be awesomer, I have ideas!)
   - other sorts of complicated ui that might involve checking what extensions a user has, etc.

Ideally we *also* have tests, but this is the sort of stuff where manual verification and faffing about is easier in a FIrefox.

## Desired solution
I want this work flow to work trivially.

1.  create an empty profile from the command line that has
    - useful sdk and debugging prefs
    - extended with any prefs of interest from the user
2.  be able to re-use this profile
    - by cloning (`cp -rp` is acceptable)
    - 'with modification'

`jpm` could solve this lots of ways:

1.  `jpm make-profile -d somedir --prefs some.flie`   AND `jpm run --profile ./somedir  --prefs some.json  --no-copy` all have to work right.  :)  I am not sure what WORK RIGHT even is, to be honest :)  Right now, some parts of that don't work.  

It might be nice to bring back the `--extensions` from `cfx`

2.  We could have a better profile creation and modificaiton tool, starting from `firefox-profile` that has all sorts of useful subcommands?
    
    - mod prefs
    - add extensions
    - localStore some things
    - fill places.db

3.  Simplest might be to 


I have been working on my own `shield-studies` CLI

kumar303:
Yeah, I feel like creating a new tool that automates the creation of a profile and populates it with the values you need is a better way to go here. The jpm codebase is crufty, full of surprises, and poorly tested. That makes it hard to introduce new features like this. 

I'm not really opposed to this patch technically (although it will need test coverage). It just feels like something you could do more easily by writing a standalone script that merges a custom profile with `user.js` to create a new profile. You could then start up jpm with that new profile.

PR #578, opened 2016-09-19
Feature: add support for hybrid SDK addons

rpl:
This PR introduces changes needed to support (and simplify) the creation of hybrid SDK addons, which will be able to embed a WebExtension inside a legacy SDK addon and exchange messages between the two contexts (SDK hybrid addons depends from [Bug 1269347 - Expose the optional embedded webextension as a builtin SDK module](https://bugzilla.mozilla.org/1269347) on the Firefox side).

This patches enables `jpm` to build correctly SDK hybrid addons by applying the following changes:
- ensuring all the files from the `webextension/` dir (the hybrid add-on sub-directory which is going to contain all the embedded webextension resources) are going to be included in the built xpi file (this change is needed because currently all the "manifest.json" files are ignored, as part of #566)
- enabling the `hasEmbeddedWebExtension` property in the generated install.rdf, if it is specified and set to true in the add-on package.json file (because most of the SDK add-ons do not create the install.rdf directly, but they use the install.rdf generated by jpm from the add-on package.json)


lib/ignore.js
```diff
@@ -56,7 +56,7 @@ function ignore(dir, options) {
     })
     .then(function(lines) {
       // Add "manifest.json" to always exclude it from the xpi.
```
kumar303: To make it extra clear, you should update this comment mentioning the embedded case.

test/unit/test.xpi.js
```diff
@@ -581,4 +584,30 @@ describe("lib/xpi", function() {
         });
       });
   });
+
+  it("Don't exclude webextension/manifest.json used in hybrid addons", function() {
```
kumar303: nit: `it("Doesn't exclude...")`

rpl:
> We'll have to somehow remember to update the manifest docs after this change.

@kumar303 did you mean package.json, am I right? 

Thanks for mention it! I'm working with @wbamberg on the planning and creation of the new MDN doc pages for the hybrid addons and the related webextensions-examples, and the [MDN page related to the package.json](https://developer.mozilla.org/en-US/Add-ons/SDK/Tools/package_json) is definitely one of the pages that are going to be in the list of the pages to update.

kumar303:
> @kumar303 did you mean package.json, am I right? 

Oops, yes I did. The doc you linked to is the one I had in mind.

rpl:
@kumar303 I've just rebased this pull request on top of the changes in the linting introduced by #581 
(just to be sure that the changes introduced by this patches still passes the linting step, before proceeding with merging it) and the build on travis is green.

kumar303:
looks good!

PR #581, opened 2016-09-20
chore: Switch from JSHint/JSCS to ESLint

### LINKED ISSUE #571: Move from jscs to ESLint?
JSCS is declared to be deprecated and merged into ESLint: https://github.com/jscs-dev/node-jscs. I haven't tried ESLint, though.

pdehaan:
Fixes #571


freaktechnik:
According to http://eslint.org/docs/user-guide/configuring#configuration-file-formats using `.eslintrc` without file ending is deprecated, wouldn't it make sense to add the file ending to ensure the config is future-proof?

Is there really a need for two targets? I only ever see `lint` used, never `eslint`.

pdehaan:
> According to http://eslint.org/docs/user-guide/configuring#configuration-file-formats using `.eslintrc` without file ending is deprecated, wouldn't it make sense to add the file ending to ensure the config is future-proof?

Good catch. I did that for the top level, but totally choked on the override config in test/.


> Is there really a need for two targets? I only ever see lint used, never eslint.

A _need_? No. But it can make things a lot easier if we add more linting tools in the future. For example, if we ever add CSS linting [or whatever], we could just tweak the package.json's `lint` task and not have to edit .travis.yml, prepush hook, pretest hook, and other places. We've used it in other projects with some success (since every project has a common "lint" task, you don't need to think about it very hard). But I'm happy to remove if it's excessive here.

package.json
```diff
@@ -61,9 +58,8 @@
     "open": "0.0.5",
     "promzard": "0.3.0",
     "read": "1.0.7",
-    "request": "2.67.0",
```
pdehaan: Removed unused deps, per https://github.com/mozilla-jetpack/jpm/pull/581#issuecomment-248170571

package.json
```diff
@@ -76,11 +72,10 @@
     "conventional-changelog-cli": "1.2.0",
     "conventional-changelog-lint": "1.0.0",
     "dive": "0.4.0",
+    "eslint": "2.13.1",
```
pdehaan: Oh, and I forgot that I should throw a comment on here...

I'm not sure what version of ESLint to use. Per http://eslint.org/docs/user-guide/migrating-to-3.0.0

> With ESLint v3.0.0, we are dropping support for Node.js versions prior to 4. Node.js 0.10 and 0.12 are in maintenance mode and Node.js 4 is the current LTS version. If you are using an older version of Node.js, we recommend upgrading to at least Node.js 4 as soon as possible. If you are unable to upgrade to Node.js 4 or higher, then we recommend continuing to use ESLint v2.x until you are ready to upgrade Node.js.

> Important: We will not be updating the ESLint v2.x versions going forward. All bug fixes and enhancements will land in ESLint v3.x.

I wasn't 100% on our support for Node < 4 (the [package.json](https://github.com/mozilla-jetpack/jpm/blob/413022d1c647565d24090061c1c5898ae092d24d/package.json#L9-L12) mentions `node>=0.10`), so I left this as **eslint@2**. If we're only supporting Node >= 4, we should probably upgrade to **eslint@3**.

freaktechnik: As travis only runs 4+, and this is a devDependency this shouldn't be a problem, there are already devDependencies that only support node 4 and up (get-firefox as an example).

pdehaan: Updated to **eslint@3** and re-pushed.

package.json
```diff
@@ -16,10 +16,10 @@
     "addon_runners": "node ./test/run.js addon_runners",
     "unit": "node ./test/run.js unit",
     "documentation": "mocha -t 5000 test/documentation",
-    "jshint": "jshint bin/jpm lib test --exclude test/addons,test/addon_runners,test/fixtures",
-    "jscs": "jscs bin/jpm lib",
+    "eslint": "eslint bin/jpm lib test",
```
kumar303: I think the `lint` command is enough, right? I'd like to avoid aliases if we don't need them.

package.json
kumar303: This is great but I'd like to make sure it doesn't happen again. Can we use [this rule](https://www.npmjs.com/package/eslint-plugin-dependencies) maybe?

kumar303:
hmm, I'm not sure what these test failures are about

pdehaan:
> hmm, I'm not sure what these test failures are about

I think this broke when I added the `eqeqeq` rule, but not sure what would have caused that. Unless it's the `return (process.env.TRAVIS == "true");`.

All the other changes in that commit were just comparing to strings or numbers. I'll see if switching back to `==` fixes the failing Travis.

Computers are hard.

package.json
pdehaan: No. I added that plugin and configured it, and then added the **request** dependency back and ESLint still passed, so it doesn't seem to do the same thing as depcheck.

We can probably just use **depcheck** either on Travis-CI only, or add it to devDependencies.

package.json
pdehaan: Removed `eslint` alias and just left `lint` (which still lets me be consistent with our other repos). :+1:

pdehaan:
Just talking to myself here while my Travis-CI is running in the background.

So, with **depcheck**, we can get it to ignore certain dependencies using the `--ignores` flag (per [usage](https://github.com/depcheck/depcheck#usage) docs).

Not super pretty though:

```sh
$ depcheck --ignores=commander,jpm-core,node-watch,object-assign,open,eslint-plugin-*,husky,release-it

Missing dependencies
* sdk
* chrome

$ echo $? # 0
```

So it doesn't seem to know how to handle the quirky `sdk` and `chrome` faux dependencies, ~~and always seems to exit with an exit code of 0 (so it won't fail travis ever, seemingly).~~ Actually, it seems the error code was fixed in the newer **depcheck@0.6.4**, but still gives an exit code of 255 for the missing `sdk` and `chrome` dependencies, so we'd need to figure that config out, or file an upstream bug.

freaktechnik:
I think the proper fix to ignore the `sdk`and `chrome` dependencies would be to disable the depcheck in data/, since that doesn't have any actual requiring in there.

package.json
kumar303: Hmm, I swear I've done this before (cant' find it) but maybe it was for a project only with ES6 imports.

kumar303: ok, if you can run depcheck as part of the Travis build, that would be great

kumar303:
The latest failure looks like a flaky timeout so I restarted it

pdehaan:
> I think the proper fix to ignore the sdkand chrome dependencies would be to disable the depcheck in data/, since that doesn't have any actual requiring in there.

That'd solve 2/3 of the problems. We'd still need a solution for the `require('chrome')` in test/fixtures/.../addon/runner.js:

> Per our [**mozilla-jetpack/jpm**](https://github.com/mozilla-jetpack/jpm) repo, we have the following files (and special dependencies):

> 1. [data/index.js:1](https://github.com/mozilla-jetpack/jpm/blob/413022d1c647565d24090061c1c5898ae092d24d/data/index.js#L1) &mdash; `var self = require("sdk/self");`
> 1. [data/test-index.js:19](https://github.com/mozilla-jetpack/jpm/blob/413022d1c647565d24090061c1c5898ae092d24d/data/test-index.js#L19) &mdash; `require("sdk/test").run(exports);`
> 1. [test/fixtures/mock-sdk/lib/sdk/addon/runner.js:5](https://github.com/mozilla-jetpack/jpm/blob/413022d1c647565d24090061c1c5898ae092d24d/test/fixtures/mock-sdk/lib/sdk/addon/runner.js#L5) &mdash; `const { Cc, Ci } = require("chrome");`

> &mdash; via depcheck/depcheck#167

pdehaan:
In this new "GitHub review" world, is there a way that I can mark the requested changes as complete? Or can only repo owners/admins/members do that?

TL;DR: Merge this quick while it's green!

kumar303:
Thanks @pdehaan !

Can you file a new issue to fix the TODOs that you commented about in `.eslintrc.yml`? I think those are all valid, I'd especially like to turn on `no-console` and `max-len`.

pdehaan:
> Can you file a new issue to fix the TODOs that you commented about in .eslintrc.yml? I think those are all valid, I'd especially like to turn on no-console and max-len.

Filed #585 which is the overall TODO issue, and submitted #587 which is the PR to fix `no-console` and `max-len`. :eyes:

Issue #585, opened 2016-09-21
Fix TODOs in .eslintrc.yml


pdehaan:
The current JSHint/JSCS => ESLint added a few TODOs to re-evaluate rules.
Now that we've landed the ESLint patch, we should reconsider the rules.

```sh
$ cat .eslintrc.yml | grep -i TODO

  brace-style: off   # TODO: Set to warn|error?
  comma-dangle: off  # TODO: Set to `[error, always|never]`?
  curly: off         # TODO: Set to error?
  max-len: [off, 80] # TODO: Set to warn|error?
  no-console: off    # TODO: Set to warn|error?
```

pdehaan:
Fixed `max-len` and `no-console` in PR #587.

PR #587, opened 2016-09-21
chore: Add ESLint no-console and max-len rules

pdehaan:
Ref #585

### DIFF
--- a/lib/firefox.js
+++ b/lib/firefox.js
@@ -13,7 +13,7 @@ var GARBAGE = [
   /\[JavaScript Warning: "TypeError: "[\w\d]+" is read-only"\]/,
   /JavaScript strict warning: /,
   /\#\#\#\!\!\! \[Child\]\[DispatchAsyncMessage\]/,
-  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/,
+  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/, // eslint-disable-line max-len
   /JavaScript strict warning: resource:\/\/gre\/components\/nsSearchService\.js/
 ];
 
--- a/lib/post.js
+++ b/lib/post.js
@@ -40,16 +40,16 @@ function postXPI(manifest, options) {
           reject(e);
         });
         client.on("close", function(hadError) {
-          console.log(hadError ? "Posting XPI to %s failed" :
+          console.log(hadError ? "Posting XPI to %s failed" : // eslint-disable-line no-console
                       "Posted XPI to %s", postURL);
           fs.unlink(xpiPath);
-          console.log("Removed XPI from " + xpiPath);
+          console.log("Removed XPI from " + xpiPath); // eslint-disable-line no-console
           resolve();
         });
       });
     });
   }).then(null, function(err) {
-    console.error(err);
+    console.error(err); // eslint-disable-line no-console
   });
 }
 module.exports = postXPI;
--- a/lib/post.js
+++ b/lib/post.js
@@ -9,6 +9,8 @@ var fs = require("fs");
 var url = require("url");
 var net = require("net");
 var when = require("when");
+var utils = require("./utils");
+var logger = utils.console;
 
 function postXPI(manifest, options) {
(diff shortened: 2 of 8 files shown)


pdehaan:
Unfortunately there isn't currently a great way to have ESLint ignore really long regular expressions, so I had to use `// eslint-disable-line max-len`.

Once a fix for https://github.com/eslint/eslint/issues/3229 lands, we can remove those shameful `eslint-disable-line`s and use the new `max-len` options instead.

lib/post.js
```diff
@@ -40,16 +40,16 @@ function postXPI(manifest, options) {
           reject(e);
         });
         client.on("close", function(hadError) {
-          console.log(hadError ? "Posting XPI to %s failed" :
+          console.log(hadError ? "Posting XPI to %s failed" : // eslint-disable-line no-console
```
pdehaan: This looks a bit weird since the `console.log()` is wrapped on 2 lines, but I had to put the comment on the same line as opening `console.log()`.

kumar303: These should be switched to the logger. You can set it up like this at the top of the module:
````
var utils = require("./utils");
var logger = utils.console;
````
And use it like either `logger.info(...)` or `logger.error(...)`

File: lib/firefox.js [hunk 2]
```diff
@@ -13,7 +13,7 @@ var GARBAGE = [
   /\[JavaScript Warning: "TypeError: "[\w\d]+" is read-only"\]/,
   /JavaScript strict warning: /,
   /\#\#\#\!\!\! \[Child\]\[DispatchAsyncMessage\]/,
-  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/,
+  /JavaScript strict warning: resource:\/\/\/modules\/sessionstore\/SessionStore\.jsm/, // eslint-disable-line max-len
```
kumar303: Can these be concatenated and broken into separate lines like `/first/ + /second/` ? That might be harder to read, I'm not sure.

lib/sign.js
```diff
@@ -57,11 +57,11 @@ function sign(options, config) {
       }
     });
     if (missingOptions.length) {
-      console.error();
+      console.error(); // eslint-disable-line no-console
```
kumar303: I think all these do need to be direct console calls. I forget why exactly but there was some reason.

pdehaan: Only reason I can think of was to add some extra leading/trailing whitespace in the console.
I was trying to figure out how to wrap this in a single `console.error()` call, but it was a confusing mix of:

```
var errors = missingOptions.map(function(flag) {
  return `  error: missing required option '${flag}'`;
});
console.error(`\n${errors.join("\n")}\n`); // eslint-disable-line no-console
```

And I wasn't convinced that my theory would be any cleaner or clearer.

lib/firefox.js
pdehaan: Can you concatenate Regular Expressions?
Do half these even need to be regular expressions vs plain strings?

kumar303: ha, well, it's legacy cruft for sure. I wouldn't worry about it too much, eslint disable comments are fine.

pdehaan: Yeah, I didn't feel like putting in TOO much effort on this, since it seems like ESLint has a fix in progress for ignoring long regular expressions.

lib/post.js
```diff
@@ -40,16 +42,19 @@ function postXPI(manifest, options) {
           reject(e);
         });
         client.on("close", function(hadError) {
-          console.log(hadError ? "Posting XPI to %s failed" :
-                      "Posted XPI to %s", postURL);
+          if (hadError) {
+            logger.error("Posting XPI to " + postURL + " failed");
+          } else {
+            logger.info("Posted XPI to " + postURL);
+          }
```
pdehaan: And I guess I should also call out that I slightly changed the logic here, where I converted this to an `if..else` instead of a ternery statement, because it let me use `logger.error()` and `logger.info()` a bit more appropriately [said the guy with zero context].

kumar303: I saw that, looks good

kumar303:
Thanks!

PR #592, opened 2016-10-04
fix: Fixed bootstrap.js and install.rdf not generating for Firefox >= 50.0a

### LINKED ISSUE #590: Will build with AOM support if min version is bigger than 50.0a
AOM support, which is install.rdf free SDK extensions never happened, but the code still thinks it will be in for Firefox 50.0a, which is defined in https://github.com/mozilla-jetpack/jpm/blob/master/lib/settings.js#L13. I think the easiest fix is to remove https://github.com/mozilla-jetpack/jpm/blob/master/lib/utils.js#L27 and use false as replacement wherever it is used.

freaktechnik:
This fixes #590 by removing every trace of that feature. I did not remove the tests since it seems to test unrelated things, like the actual generation of install.rdf and bootstrap.js

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -52,12 +52,7 @@ program
   .option("--debug", "Enable the add-on debugger when running the add-on")
   .option("--check-memory",
           "Enable leaked tracker that attempts to report compartments leaked")
-  .option("--profile-memory", "Enable profiling of memory usage")
-  .option(
-    "--retro",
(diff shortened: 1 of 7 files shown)


kumar303:
more-red-than-green-award

atsay:
Well I can't really beat the giant red ribbon, but I did add this contribution to the [recognition page](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#October_2016) :-)

Issue #446, opened 2015-12-21
Please support http proxy servers for jpm. 


jnurthen:
jpm sign and other such commands seem intended to support corporate users who are normally behind proxy servers. As such jpm needs to support their configurations.

kumar303:
Do you think simply specifying a proxy server [to request](https://github.com/request/request#proxies) would work for your case? Could you try editing [this line](https://github.com/mozilla-jetpack/jpm/blob/master/lib/amo-client.js#L365) to see if it works for you? I'm guessing something like:
````
requestConf.proxy = 'https://<your-proxy>';
````
So, in other words, I'm not asking for a patch but just a proof to see if the `proxy` option works for what you need.

kumar303:
and you could run it with `jpm --verbose sign ...` to get more info

jnurthen:
That seemed to work for the upload great. I set
requestConf.proxy = 'http://myproxy:proxyport'; 
and the upload completed.

I got a timeout error eventually but it may have been that the addon took a while to complete the validation. It was available on MDN so the upload part completed ok.

kumar303:
ok, thanks, we will add a `--proxy` command line option for this.

jnurthen:
Thanks - that would be great.

kumar303:
This depends on https://github.com/mozilla-jetpack/jpm/issues/519 and https://github.com/mozilla/sign-addon/issues/49

ingoe:
+1

vanthoom:
Hi whats the current status and plan to suport this?

kumar303:
No plans yet. I can mentor a patch to sign-addon for it. This is step 1.

kumar303:
I was making other changes to `sign-addon` so I started a patch for proxy support: https://github.com/mozilla/sign-addon/pull/120 On the JPM side, we will need more work thought, such as https://github.com/mozilla-jetpack/jpm/issues/519

kumar303:
The new `--api-proxy` option is available now in jpm 1.3.0 https://www.npmjs.com/package/jpm

@ingoe thanks again for coming back to this patch!

vvijayabhanu:
Hi

Is this apiproxy fields support a proxy machine that requires username and password

PR #593, opened 2016-10-06
fix: Add support for proxy configuration

ingoe:
Allows using a proxy when signing addons.
Since mozilla/sign-addon already supports a proxy configuration, this change will pass on the option to sign-addon

Fixes https://github.com/mozilla-jetpack/jpm/issues/446

### DIFF
--- a/bin/jpm
+++ b/bin/jpm
@@ -101,6 +101,7 @@ program
           "optional XPI to be signed. By default, an XPI will be built " +
           "from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
+  .option("--api-proxy <url>", "Optional proxy to use for all API requests")
   .action(function(options) {
     signCmd(program, options);
   });
--- a/bin/jpm
+++ b/bin/jpm
@@ -101,7 +101,8 @@ program
           "optional XPI to be signed. By default, an XPI will be built " +
           "from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
-  .option("--api-proxy <url>", "Optional proxy to use for all API requests")
+  .option("--api-proxy <url>", "Optional proxy to use for all API requests." +
+          " Example: https://yourproxy:6000")
   .action(function(options) {
     signCmd(program, options);
   });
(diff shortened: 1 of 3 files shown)


kumar303:
Hi @ingoe, do you still have time to make the requested changes? Thanks again for working on this.

bin/jpm
```diff
@@ -101,6 +101,7 @@ program
           "optional XPI to be signed. By default, an XPI will be built " +
           "from your working directory")
   .option("--timeout <milliseconds>", "time to wait for a signing result")
+  .option("--api-proxy <url>", "Optional proxy to use for all API requests")
```
kumar303: I think it would be helpful add a second sentence to this help message that reads: "Example: https://yourproxy:6000"

test/unit/test.sign.js
```diff
@@ -173,6 +173,20 @@ describe("sign", function() {
     }).catch(done);
   });
 
+  it("can configure a proxy", function(done) {
+    runSignCmd({
```
kumar303: I know another test you probably modeled this on didn't do so, but it's easier to just return promises. Your test would change to the following. Notice how you have to drop the `done` argument in the callback for this to work.
````js
it("can configure a proxy", function() {
  return runSignCmd({
    ...
  }).then(function() {
    expect(signAddonCall.call[0].apiProxy)
      .to.be.equal("some-proxy");
  });
});
````

File: lib/sign.js [hunk 3]
```diff
@@ -119,6 +119,7 @@ function sign(options, config) {
       apiUrlPrefix: options.apiUrlPrefix,
       verbose: options.verbose,
       timeout: options.timeout || undefined,
+      apiProxy: options.apiProxy || undefined
```
kumar303: could you group this with all of the other `api*` prefixed options to make the list easier to scan?

test/unit/test.sign.js
```diff
@@ -173,6 +173,19 @@ describe("sign", function() {
     }).catch(done);
   });
 
+  it("can configure a proxy", function() {
+    runSignCmd({
```
kumar303: Almost there. You need to return this value so that mocha will know to fail when the promise fails.

caitmuenster:
Hey @ingoe! Welcome onboard! Your contribution has been noted on our [recognition wiki](https://wiki.mozilla.org/Add-ons/Contribute/Recognition#February_2017). 

Looking forward to seeing you around!

PR #599, opened 2016-10-11
chore: Don't check readme pre push but check commit message

### LINKED ISSUE #594: Run changelog lint pre-push?
I think running the changelog lint pre-push would make sense. Maybe some of the current checks could be removed, not sure how important it is to check the spelling of the README on every push.

freaktechnik:
This runs the changelog-lint script at the git "commit-msg" hook, like the web-ext setup does. I've also tested it on Windows, where it works, except with GitHub for Windows, which probably doesn't support git hooks, though I suspect push would fail there too.

Further it removes the documentation spellcheck from pre-push. While doing that, I was wondering, if ti was worth adding `--cache` to the pre-push eslint, though I think there's not enough push frequency and it's not that slow to warrant that.

This fixes #594

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,8 @@
     "documentation": "mocha -t 5000 test/documentation",
     "lint": "eslint bin/jpm lib test",
     "test": "node ./test/run.js",
-    "prepush": "npm run lint && npm run documentation",
+    "prepush": "npm run lint",
+    "commitmsg": "npm run changelog-lint",
     "changelog": "conventional-changelog -p angular -u",
     "changelog-lint": "conventional-changelog-lint --from master",
     "get-unbranded-firefox": "get-firefox -ecb unbranded-release"
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lint": "eslint bin/jpm lib test",
     "test": "node ./test/run.js",
     "prepush": "npm run lint",
-    "commitmsg": "npm run changelog-lint",
+    "commitmsg": "cat \"$GIT_PARAMS\" | conventional-changelog-lint",
     "changelog": "conventional-changelog -p angular -u",
     "changelog-lint": "conventional-changelog-lint --from master",
     "get-unbranded-firefox": "get-firefox -ecb unbranded-release"
@@ -76,7 +76,7 @@
     "eslint-plugin-dependencies": "1.3.0",
     "get-firefox": "1.5.0",
     "glob": "5.0.3",
-    "husky": "^0.10.1",
+    "husky": "0.11.9",
     "mocha": "2.5.3",
     "release-it": "2.3.1",
     "rimraf": "2.3.2",


kumar303:
Honestly, I think we can remove the eslint check from the pre-push hook altogether. The main reason I wanted to get the changelog-lint as a commit hook is because once you push a bad commit message, the PR will fail indefinitely and this is hard to fix if you aren't familiar with rebasing.

freaktechnik:
@kumar303 just tested this version in a windows shell with git and it worked fine, though I still had an error when using the GitHub client.

kumar303:
Huh, this is confusing. On Mac OS X (tried in Node 6.3.0 and 4.3.0), I see this:

````
$ git commit -a -m 'nope'

> jpm@1.2.2 commitmsg /Users/kumar/dev/jpm
> cat "$GIT_PARAMS" | conventional-changelog-lint

/Users/kumar/dev/jpm/node_modules/conventional-changelog-lint/distribution/cli.js:6
let main = function () {
^^^
SyntaxError: Unexpected strict mode reserved word
    at exports.runInThisContext (vm.js:73:16)
    at Module._compile (module.js:443:25)
    at Object.Module._extensions..js (module.js:478:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:501:10)
    at startup (node.js:129:16)
    at node.js:814:3
````

However, if I run this in my shell it works as expected:
````
$ cat .git/COMMIT_EDITMSG | conventional-changelog-lint
⧗   input: nope
✖   message may not be empty [subject-empty]
✖   type may not be empty [type-empty]
✖   type must be one of ["chore", "docs", "feat", "fix", "perf", "refactor", "style", "test"] [type-enum]
✖   found 3 problems, 0 warnings
````
I don't understand what black magic npm (or husky) might be doing to cause this. On a hunch, I tried applying this patch but it made no difference:

````diff
diff --git a/package.json b/package.json
index 0c9a56f..511c0c6 100644
--- a/package.json
+++ b/package.json
@@ -19,9 +19,10 @@
     "lint": "eslint bin/jpm lib test",
     "test": "node ./test/run.js",
     "prepush": "npm run lint",
-    "commitmsg": "echo \"params: $GIT_PARAMS\"",
+    "commitmsg": "cat \"$GIT_PARAMS\" | npm run changelog-lint-from-stdin --silent",
     "changelog": "conventional-changelog -p angular -u",
     "changelog-lint": "conventional-changelog-lint --from master",
+    "changelog-lint-from-stdin": "conventional-changelog-lint",
     "get-unbranded-firefox": "get-firefox -ecb unbranded-release"
   },
   "homepage": "https://github.com/mozilla-jetpack/jpm",
````
Any ideas on what could be wrong? I take it that you do not see this on your system?

freaktechnik:
The important bit is updating the husky version, too and then regenerating the git hooks, since the current version the package.json specifies does not have $GIT_PARAMS in the hook script template yet.

kumar303:
I removed my entire `node_modules` and re-installed a couple times (when trying it each Node version) but that didn't seem to help

PR #602, opened 2016-11-15
Drop node 5 build

wagnerand:


wagnerand:
@kumar303 r?

kumar303:
Good idea, thanks! For anyone else wondering why we're doing this, it's because Node 5 is not a LTS (long term support) release: https://github.com/nodejs/LTS

